PR #5, opened 2016-08-25
Validate metadata info before writing to file.

pasinskim:


writer/writer_test.go
```diff
@@ -15,17 +15,43 @@
...
 import (
+	"fmt"
 	"testing"
 
 	"github.com/mendersoftware/artifacts/metadata"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestMarshallInfo(t *testing.T) {
+	info := metadata.MetadataInfo{
```
maciejmrowiec: try out table tests, makes those easier to follow

pasinskim: Will do. I am adding things case by case, and when there is enough to add table tests I am trying to do so. Thanks for review.

PR #7, opened 2016-08-29
Introduce codecov to travis.

maciejmrowiec:


maciejmrowiec:
@pasinskim

maciejmrowiec:
kept both for now to compare, did not spend time on adding coverage limits as they require config file in the repo (something we can config later is we decide we like it)

PR #8, opened 2016-08-29
Working artifacts writer

pasinskim:
Validate artifacts layout and content.


metadata/metadata.go
```diff
@@ -53,3 +56,70 @@ func (m MetadataHeaderInfo) Validate() error {
...
+}
+
+var ErrInvalidMetadata = errors.New("invalid metadata")
+
+type Metadata struct {
+	// we don't know exactly what tyoe of data we will have here
+	data map[string]interface{}
+}
+
+func (m Metadata) Validate() error {
```
maciejmrowiec: check out https://github.com/asaskevich/govalidator

metadata/metadata.go
```diff
@@ -14,19 +14,28 @@
...
 	Format  string `json:"format"`
-	Version string `json:"version"`
+	Version int    `json:"version"`
 }
 
-type MetadataInfoJSON string
-
 func (m MetadataInfo) Validate() error {
-	if len(m.Format) == 0 || len(m.Version) == 0 {
+	if len(m.Format) == 0 || m.Version == 0 {
```
kacf: Maybe you should rather verify that:

```
if m.Format != "mender" || m.Version != 1
```

metadata/metadata.go
```diff
@@ -53,3 +62,148 @@ func (m MetadataHeaderInfo) Validate() error {
...
+	if len(m.Rootfs) == 0 {
+		return ErrInvalidTypeInfo
+	}
+	return nil
+}
+
+var ErrInvalidMetadata = errors.New("invalid metadata")
+
+type Metadata struct {
+	// we don't know exactly what tyoe of data we will have here
```
kacf: `s/tyoe/type/`

metadata/metadata.go
```diff
@@ -53,3 +62,148 @@ func (m MetadataHeaderInfo) Validate() error {
...
+	if deviceType == nil || imageID == nil {
+		return ErrInvalidMetadata
+	}
+	return nil
+}
+
+var ErrInvalidFilesInfo = errors.New("invalid files info")
+
+type MetadataFile struct {
+	File string `json:"file"`
```
kacf: Hmm, is this correct? According to the document, each element of the list in `files` is a bare string, not a key/value pair whose key is `file`.

metadata/metadata_test.go
```diff
@@ -54,3 +60,246 @@ func TestValidateHeaderInfo(t *testing.T) {
...
+}
+
+var dirStructOK = []MetadataDirEntry{
+	{Path: "files", IsDir: false},
+	{Path: "type-info", IsDir: false},
+	{Path: "meta-data", IsDir: false},
+	{Path: "checksums", IsDir: true},
+	{Path: "checksums/image.sha", IsDir: false},
+	{Path: "signatures", IsDir: true},
+	{Path: "signatures/iamge.sig", IsDir: false},
```
kacf: You've misspelled `image` as `iamge` in all the test cases. This would actually be an error, so you could consider testing for this. All the checksums and signatures must have a 1-to-1 name relationship with all the payload files.

metadata/metadata.go
```diff
@@ -14,19 +14,28 @@
...
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/mendersoftware/log"
+)
+
+type Validater interface {
```
kacf: `s/Validater/Validator/`

writer/writer.go
```diff
@@ -15,27 +15,380 @@
...
+	hex.Encode(upd.checksum, h.Sum(nil))
+	log.Debugf("hash of file: %v (%x)\n", upd.path, upd.checksum)
+	return nil
+}
+
+func (av ArtifactsWriter) getHeaderInfo(updates []os.FileInfo) metadata.MetadataHeaderInfo {
+	// for now we have ONLY one type of update - rootfs-image
+	headerInfo := metadata.MetadataHeaderInfo{}
+
+	// TODO: should we store update name as well?
```
kacf: I think this can be covered either in the `type-info` or `meta-data` file.

writer/writer.go
```diff
@@ -15,27 +15,380 @@
...
+		defer gz.Close()
+		tw = tar.NewWriter(gz)
+	} else {
+		tw = tar.NewWriter(destination)
+	}
+	defer func() {
+		// make sure to check the error on Close
+		if err := tw.Close(); err != nil {
+			log.Errorf("artifacts writer: error closing archive writer")
+			panic(err)
```
kacf: Is `panic()` really a good idea? I'm thinking it might happen in cases like disk full for example, although I think that might be revealed earlier than on `Close()`. I'm not sure though. But `panic()` might be a bit excessive in that case.

writer/writer.go
```diff
@@ -15,27 +15,380 @@
...
+	return av.writeArchive(archive, hCnt, true)
+}
+
+func (av *ArtifactsWriter) removeCompressedHeader() error {
+	// remove temporary header file
+	return os.Remove(filepath.Join(av.updateLocation, "header.tar.gz"))
+}
+
+func (av *ArtifactsWriter) createArtifact(files []os.FileInfo) error {
+	artifact, err := os.Create(filepath.Join(av.updateLocation, "artifact.mender"))
```
kacf: I suppose the name `artifact.mender` should be variable, no? So you can have `beaglebone-update-2016-Q1.mender`, for example.

writer/writer.go
```diff
@@ -15,27 +15,380 @@
...
+			filepath.Join(av.updateLocation, location.Name()))
+		if err != nil {
+			return err
+		}
+		if err = av.processUpdateBucket(location.Name()); err != nil {
+			return err
+		}
+	}
+	// create compressed header; the intermediate step is needed as we
+	// can not create tar archive containing files compressed on the fly
```
kacf: Is this a limitation of the tar/gz interfaces? I was hoping they would both accept streams.

writer/writer_test.go
```diff
@@ -25,33 +30,210 @@ import (
...
+	{Path: "0000/scripts/check", IsDir: true},
+}
+
+var dirStructOKAfterWriting = map[string]metadata.MetadataDirEntry{
+	".":                               {Path: ".", IsDir: true, Required: true},
+	"data":                            {Path: "data", IsDir: true, Required: true},
+	"data/0000.tar.gz":                {Path: "data", IsDir: false, Required: true},
+	"0000/data":                       {Path: "0000/data", IsDir: true, Required: true},
+	"0000/data/update.ext4":           {Path: "0000/data/update.ext4", IsDir: false, Required: true},
+	"0000/data/update_next.ext3":      {Path: "0000/data/update_next.ext3", IsDir: false, Required: true},
```
kacf: These (the `0000/data/*` ones) shouldn't end up in the artifact file. Are they just for testing the intermediary step?

kacf:
Done reviewing! I admit I wasn't able to follow everything, but I will do additional checking on the output of this when we start integrating it into the build.

metadata/metadata.go
pasinskim: Not here. This is supposed to be "general" check to validate content. We will check for exact values while we will read this in client.

metadata/metadata.go
pasinskim: Fixed!

metadata/metadata.go
pasinskim: Right. I think we need to change documentation. It states:
```
[
  "core-image-minimal-201608110900.ext4"
]
```
but this is not valid JSON.

metadata/metadata_test.go
pasinskim: Fixing misspelling. Actually metadata is quite basic check. This will be checked (and actually is) in writer part of library.

metadata/metadata.go
pasinskim: Go convention is quite strict about this. If you have an interface with only one method its name should be `method name + (e)r` so `Validater` in this case.

writer/writer.go
pasinskim: I have been more thinking that instead of: 
```
{
  "updates": [
    {
      "type": "rootfs-image"
    },
    {
      "type": "rootfs-image"
    },
    {
      ...
    }
  ]
}
```
we can have:
```
{
  "updates": [
    {
      "type": "rootfs-image"
      "update": "0000/core-image-minimal-201608110900.ext4"
    },
    {
      "type": "rootfs-image"
      "update": "0000/core-image-base-201609010900.ext3"
    },
    {
      ...
    }
  ]
}
```
Having that maybe we can get rid of some other files (like files), or combine files and type-info.

writer/writer.go
pasinskim: I've been thinking about this as well. It might be tricky to do something else. 
Anyway, this is "just" writing part of library. So if you can not write artifact file you simply panic and then you will try once more. 
I am open for suggestions here.

writer/writer.go
pasinskim: If you say so... I will add this as a parameter then.

writer/writer.go
pasinskim: They are, and in many cases we are doing thing on the fly. See `writeArchive()`. Here we are writing a file/stream to gzip (using `io.Copy()`) which is then written to tar which is written to file. 
Here the limitation is that we need to write tar header before writing content. And part of header is size of file which will be written. So we need to have a file in advance to check size. 
Maybe there is some work-around possible, but since this is ONLY for writing compressed headers (pretty small) and will be done in Yocto only I don't want to spend time fixing it.
FYI reader won't store any intermediate files. It will read everything on the fly.

writer/writer_test.go
pasinskim: Not really. I think we have agreed on the structure of the directory BEFORE packing it to artifact archive. Data is something that needs to be there. We are not removing it while creating archive. One reason is that we might need to redo the whole thing in case of some errors.

metadata/metadata.go
kacf: Ack

metadata/metadata.go
kacf: Aha, my bad then! You are proposing:

```
[
  "file": "core-image-minimal-201608110900.ext4"
]
```

is that right? Looks good to me at least. Make sure you update the format document at https://github.com/mendersoftware/meta-mender/blob/master/Documentation/artifact-format.md as well.

metadata/metadata_test.go
kacf: Got it!

metadata/metadata.go
kacf: Ugh, that's a bizarre requirement, it enforces spelling errors! But if that's how it is, we have no choice I guess.

writer/writer.go
kacf: No, there is a specific reason I avoided this: An update is not guaranteed to have just one payload file, and some might not have any at all. Some updates may provide a URL instead, or login details somewhere, and this will be in the `type-info` header. In order to keep the format modular, when reading the `header-info` file you should not need to know anything about the expected contents of an update, only what type it is. This allows you to skip or ignore update types you don't know about.

We could combine `files` and `type-info` (they were combined in the first draft actually). The reason I split them is to (hopefully) make the logic simpler across different update type implementations: `files` will usually be the same format in most update types, hence can share code, whereas `type-info` is very likely to be different. Not sure if this is a great reason, but at least the content of `files` should be inside the `headers` folder, not in the `header-info` header itself.

writer/writer.go
kacf: I can see how it's difficult in the `defer` context. Here's my suggestion:

```
	defer func() {
		if tw {
			// Ignore error since we cannot do anything about it here.
			tw.Close()
		}
	}()

... // Later, at the end of the function, outside defer.

	// make sure to check the error on Close
	defer tw = nil // <-- To prevent the first defer from also doing Close.
	if err := tw.Close(); err != nil {
		return errors.Wrapf(err, "artifacts writer: error closing archive writer")
	}
	return nil
}
```

writer/writer_test.go
kacf: Ack.

kacf:
One thing to point out as well, you've used gz throughout, but we talked about using xz. Was this deliberate because it's not supported / harder to implement / etc, or did you pick the wrong one?

pasinskim:
I've picked up gzip as it has native Go support. I will continue with this one to have something by the end of the sprint. Later on (hopefully this sprint) I will do a research on xz. If I will find a nice library replacing gzip with xz should be as easy as changing few lines of code.

writer/writer.go
pasinskim: Looks like a real hack. Maybe instead I will just call close before returning from all possible places.

kacf: I guess it does a bit. :-) But it has the advantage that you still get `defer` semantics, not C semantics.

PR #13, opened 2016-09-09
Full implementation of artifacts library

pasinskim:
`writer` implements writing part. It scans directories in provided location and looks for correct update directories (each needs to be of `sprintf("%04d")` format and there must be `meta-data`, `type-info` and `data/[update_name]` files).

Based on update type (specified in `type-info`) correct parser is chosen  to parse given type of update.
This design will allow extending library and add new parsers while there will be a need for doing so.
So far we have `rootfs-image` parser implemented, for parsing artifact file as described here: https://github.com/mendersoftware/meta-mender/blob/master/Documentation/artifact-format.md#files

All parser management is done by `parser` module. Specifically there is a `Parsers` structure managing registering and iterating over parsers while scanning updates.
Parsers MUST implement `ArtifactParser` interface which provides methods for writing and reading given type of update.

`reader` is the reader implementation. It provides reading whole artifact on the fly without storing any intermediate files. Also all operations are done using `io.Copy` calls which use `io.Reader` and 'io.Writer` interfaces underneath making sure that copying is performance and memory optimized. 

Similar as `writer`, `reader` uses parsers for parsing given update type.

`archiver` package provides convenient way of storing different data types into tar archive. So far we can store plain files, binary data and all JSON convertible data structures implemented in `metadata` package. 

`metadata` provides implementation of data structures shared between the `reader` and the `writer`. All the data structures in `metadata` are implementing `Validater` interface which in turn implements `io.Writer` interface. Together with `metadata_archiver` all the `metadata` data structures are implementing both `io.Reader` and `io.Writer` for convenient usage with `io.Copy`. 

Last, `reader` and `writer` are separated to limit the client binary size, as client needs only `reader` part.


archiver/stream_archiver.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (str *StreamArchiver) Archive(tw *tar.Writer) error {
+
+	hdr := &tar.Header{
+		Name: str.name,
+		Mode: 0600,
+		Size: int64(len(str.data)),
+	}
+	log.Debugf("arch: have header: %v", hdr)
```
maciejmrowiec: I'm not sure if a stand alone library should be allowed to write logs - not something I would expect to happen.
This forces user to use logging and a specific one, I personally would be leaning towards leaving that to the user outside if and how he want to do this.

maciejmrowiec: service uses context specific logger instances which would not work well with this

metadata/metadata.go
```diff
@@ -47,6 +50,25 @@ func (i Info) Validate() error {
...
+		if err := dec.Decode(data); err != io.EOF {
+			break
+		} else if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (i *Info) Write(p []byte) (n int, err error) {
```
maciejmrowiec: Why using named returns if not using them?

metadata/metadata.go
```diff
@@ -72,42 +94,55 @@ func (hi HeaderInfo) Validate() error {
...
 // Validate validates corectness of TypeInfo.
 func (ti TypeInfo) Validate() error {
-	if len(ti.Rootfs) == 0 {
+	if len(ti.Type) == 0 {
 		return ErrValidatingData
 	}
 	return nil
 }
 
+func (ti *TypeInfo) Write(p []byte) (n int, err error) {
```
maciejmrowiec: what is `p` ?

metadata/metadata.go
```diff
@@ -72,42 +94,55 @@ func (hi HeaderInfo) Validate() error {
...
 	var deviceType interface{}
 	var imageID interface{}
 
-	for k, v := range m.data {
+	for k, v := range m {
 		if v == nil {
+			log.Error("empty element while validating metedata")
 			return ErrValidatingData
 		}
 		if strings.Compare(k, "DeviceType") == 0 {
```
maciejmrowiec: if this is a JSON key, probably would be more natural to lean towards lower case snake style

parser/parser_manager.go
```diff
@@ -0,0 +1,82 @@
...
+package parser
+
+import (
+	"archive/tar"
+	"errors"
+	"io"
+)
+
+type ArtifactParser interface {
+	ParseHeader(tr *tar.Reader, hPath string) error
```
maciejmrowiec: not sure if this is useful but I can see here reader/writer interfaces that would be composed into parser:

```
type ArticaftParseReader interface {
ParseHeader(tr *tar.Reader, hPath string) error
ParseData(r io.Reader) error
}

type ArticaftParseWriter interface {
ArchiveHeader(tw *tar.Writer, srcDir, dstDir string) error
ArchiveData(tw *tar.Writer, srcDir, dst string) error
}

type ArtifactParser interface {
ArticaftParseWriter
ArticaftParseReader
NeedsDataFile() bool
}
```

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+	files    metadata.Files
+	tInfo    metadata.TypeInfo
+	metadata metadata.Metadata
+	updates  map[string]rootfsFile
+
+	sStore string
+	dStore io.Writer
+}
+
+func NewRootfsParser(sStoreDir string, w io.Writer) RootfsParser {
```
maciejmrowiec: any particular reason why we pass a copy instead of reference ?

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+	defer tw.Close()
+
+	for _, data := range rp.updates {
+		log.Infof("processing data file: %v [%v]", data.path, data.name)
+		a := archiver.NewFileArchiver(data.path, data.name)
+		if err := a.Archive(dtw); err != nil {
+			return err
+		}
+	}
+	//TODO
```
maciejmrowiec: leftover todo

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+		dStore:  w,
+		updates: map[string]rootfsFile{}}
+}
+
+func (rp RootfsParser) NeedsDataFile() bool {
+	return true
+}
+
+func (rp *RootfsParser) ArchiveData(tw *tar.Writer, srcDir, dst string) error {
+	f, err := os.Create("/tmp/my_data.tar.gz")
```
maciejmrowiec: Why writing to tmpfs ? isn't this equivalent as keeping this in memory ?
What with about concurrent access if we have 2 instances of RootfsParser working on something?

For intermediate files this is useful: https://golang.org/pkg/io/ioutil/#TempFile

archiver/stream_archiver.go
pasinskim: Makes sense! I will remove logs.

metadata/metadata.go
pasinskim: Looks like when you are trying to implement `io.Writer` interface go insist on using some special function signatures. Must be as is.

metadata/metadata.go
pasinskim: See above. Must use a signature Go is proposing.

metadata/metadata.go
pasinskim: I've been following this: https://github.com/mendersoftware/meta-mender/blob/master/Documentation/artifact-format.md#files

parser/parser_manager.go
pasinskim: Possibly, yes. But I want to make sure that you are providing both reader and writer to be used with this library.

parser/rootfs_parser.go
pasinskim: Some old merge crap.

parser/rootfs_parser.go
pasinskim: Like above. Merge issue.

parser/rootfs_parser.go
pasinskim: If you mean return, then no reason. I guess I can change to be more consistent.

pasinskim:
@maciejmrowiec thanks for review. As this code is quite complicated (I think we might have over-engineered design of artifact file) I suggest leaving the logs for a while. 
Once we will integrate with other parts and we will see this working I will remove logs. For now those might be useful to fix some possible issues. And since those are here anyway, there is no difference if I will remove it now or in a week or so.

What do you think?

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+
+	for _, data := range rp.updates {
+		log.Infof("processing data file: %v [%v]", data.path, data.name)
+		a := archiver.NewFileArchiver(data.path, data.name)
+		if err := a.Archive(dtw); err != nil {
+			return err
+		}
+	}
+	//TODO
+	dtw.Close()
```
maciejmrowiec: it's already defered

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+		return errors.Wrapf(err, "parser: can not store meta-data")
+	}
+
+	if err := archiveChecksums(tw, updFiles,
+		filepath.Join(srcDir, "data"),
+		filepath.Join(dstDir, "checksums")); err != nil {
+		return err
+	}
+
+	//TODO: get rid of bad Joins
```
maciejmrowiec: leftover todo

parser/rootfs_parser.go
```diff
@@ -0,0 +1,346 @@
...
+
+		relPath, err := filepath.Rel(hPath, hdr.Name)
+		if err != nil {
+			return err
+		}
+
+		log.Infof("processing rootfs image header file: %v %v", hdr.Name, hdr.Linkname)
+
+		switch {
+		case i == 0 && strings.Compare(relPath, "files") == 0:
```
maciejmrowiec: may be a good idea to put fixed indexes in `const` (even a local one) to label what they mean

maciejmrowiec:
Catch with logging is that once client or server use a version with those, it's going to vendor this, so removal of logs will go away with next dependency update whenever that may be.

maciejmrowiec:
nice to have: godoc badge (so it's easy to find documentation)

pasinskim:
@maciejmrowiec makes sense. I will remove logging then. 
What about the overall functionality? Do you think it is OK?

maciejmrowiec:
it's slightly difficult to figure that out from diff :/
I know this may be a stretch but would be very helpful to get basic usage examples somewhere in doc (top section of the godoc is the part of the comments above `package xyz`) or readme which is quite common to have.

pasinskim:
This is outdated.

PR #14, opened 2016-09-14
Library

pasinskim:
`writer` implements writing part. It scans directories in provided location and looks for correct update directories (each needs to be of `sprintf("%04d")` format and there must be `meta-data`, `type-info` and `data/[update_name]` files).

Based on update type (specified in `type-info`) correct parser is chosen  to parse given type of update.
This design will allow extending library and add new parsers while there will be a need for doing so.
So far we have `rootfs-image` parser implemented, for parsing artifact file as described here: https://github.com/mendersoftware/meta-mender/blob/master/Documentation/artifact-format.md#files

All parser management is done by `parser` module. Specifically there is a `Parsers` structure managing registering and iterating over parsers while scanning updates.
Parsers MUST implement `ArtifactParser` interface which provides methods for writing and reading given type of update.

`reader` is the reader implementation. It provides reading whole artifact on the fly without storing any intermediate files. Also all operations are done using `io.Copy` calls which use `io.Reader` and 'io.Writer` interfaces underneath making sure that copying is performance and memory optimized. 

Similar as `writer`, `reader` uses parsers for parsing given update type.

`archiver` package provides convenient way of storing different data types into tar archive. So far we can store plain files, binary data and all JSON convertible data structures implemented in `metadata` package. 

`metadata` provides implementation of data structures shared between the `reader` and the `writer`. All the data structures in `metadata` are implementing `Validater` interface which in turn implements `io.Writer` interface. Together with `metadata_archiver` all the `metadata` data structures are implementing both `io.Reader` and `io.Writer` for convenient usage with `io.Copy`. 

Last, `reader` and `writer` are separated to limit the client binary size, as client needs only `reader` part.


pasinskim:
@maciejmrowiec this should be more or less done right now. I assume they might be some changes, but I made sure it will be easy to use. 
Will add some documentation in subsequent PRs.

archiver/metadata_archiver.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/mendersoftware/artifacts/metadata"
+)
+
+// NewMetadataArchiver creates streamArchiver used for storing metadata elements
+// inside tar archive.
+// data is the data structure implementing Validater interface and must be
+// a struct that can be converted to JSON (see getJSON below)
+// archivePath is the relatve path inside the archive (see tar.Header.Name)
+func NewMetadataArchiver(data metadata.Validater, archivePath string) *StreamArchiver {
```
bboozzoo: s/Validater/Validator?

archiver/stream_archiver.go
```diff
@@ -0,0 +1,57 @@
...
+	"archive/tar"
+	"bytes"
+	"io"
+
+	"github.com/pkg/errors"
+)
+
+type StreamArchiver struct {
+	archPath string
+	*bytes.Reader
```
bboozzoo: Is there a need to embed `bytes.Reader`? I understood that `StreamArchiver` is supposed to write/archive not read.

metadata/metadata_test.go
```diff
@@ -83,15 +83,15 @@ func TestValidateMetadata(t *testing.T) {
...
-		{Metadata{map[string]interface{}{}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"": nil}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"key": nil}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"key": "val"}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"DeviceType": "type"}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"DeviceType": nil, "ImageID": "image"}}, ErrValidatingData},
-		{Metadata{map[string]interface{}{"DeviceType": "device", "ImageID": "image"}}, nil},
-		{Metadata{map[string]interface{}{"DeviceType": "device", "ImageID": "image", "Data": "data"}}, nil},
+		{make(map[string]interface{}), ErrValidatingData},
+		{map[string]interface{}{}, ErrValidatingData},
```
bboozzoo: There's an alias `type Metadata map[string]interface{}` above, why not use it here?

mchalski:
(I can't pin this comment to a specific commit - guess this part was commited earlier).
can we have the well-known metadata fields factored out as constants?

I'm thinking of stuff like "DeviceType", "ImageID" (metadata.go:145). MEN-574 says we should also pull the rootfs checksum from the meta, so that's another candidate. since Metadata is a raw map, it will be useful to have it as an instant reference of what keys can be expected there. currently the only identifiers are embedded in the Metadata.Validate() func.

does it make sense? feel free to correct me if it doesn't.

archiver/metadata_archiver.go
pasinskim: This one had just a one function before `Vaalidate()` and according to naming convention it should be `function_name + er`. Right now, there is other interface there, `io.Writer`, so I will rename to `WriteValidator`.

archiver/stream_archiver.go
pasinskim: Right, but take a look at `Archive(tw *tar.Writer)` function signature. It takes `Writter` which is the place where you are writing data to. You need also `Reader` to read it from in order to be able to `io.Copy` to `Writer`. So `Reader` is used here to copy data from stream to archive `tar.Writer`.

bboozzoo:
On Mon, Sep 19, 2016 at 9:04 AM, pasinskim <notifications@github.com> wrote:

> *@pasinskim* commented on this pull request.
> ------------------------------
>
> In archiver/metadata_archiver.go
> <https://github.com/mendersoftware/artifacts/pull/14>:
>
> > +package archiver
> +
> +import (
> +	"bytes"
> +	"encoding/json"
> +	"errors"
> +
> +	"github.com/mendersoftware/artifacts/metadata"
> +)
> +
> +// NewMetadataArchiver creates streamArchiver used for storing metadata elements
> +// inside tar archive.
> +// data is the data structure implementing Validater interface and must be
> +// a struct that can be converted to JSON (see getJSON below)
> +// archivePath is the relatve path inside the archive (see tar.Header.Name)
> +func NewMetadataArchiver(data metadata.Validater, archivePath string) *StreamArchiver {
>
> This one had just a one function before Vaalidate() and according to
> naming convention it should be function_name + er. Right now, there is
> other interface there, io.Writer, so I will rename to WriteValidator.
>
The naming 'convention' was discussed a couple of times on go-nuts. Each
time using proper English words is preferable to just blindly slapping er
suffix to construct a noun, see:
- https://groups.google.com/forum/#!topic/golang-nuts/8xnggK1wo0U
- https://github.com/golang/go/issues/5421

metadata/metadata_test.go
pasinskim: Thanks! I will fix it.

pasinskim:
@bboozzoo thanks for putting some light on this naming convention stuff. Makes a lot of sense to me!

pasinskim:
@mchalski makes a lot of sense, but I didn't find any good way of doing this. It seems that it is hard to have a working JSON in Go (this is also conceptually not supported AFAIK) that you have some mandatory fields and the rest which is totally random. There are some concepts like mandatory and optional fields, but in this case you need to name all optional fields. 
Since we don't know what optional fields are we can not create structure with mix of required and whatever else is there:

```
type Metadata struct {
  DeviceType string
  ImageID string
  optional map[string]interface{}
}
```

Unless you mean something specific how to export or handle this I don't exactly know how to handle this (important IMHO) case.

pasinskim:
@mchalski @bboozzoo @maciejmrowiec after thinking a bit more about required and optional metadata I came up with this solution:
```
type Metadata struct {
	Required RequiredMetadata
	Optional optionalMetadata
}
type RequiredMetadata struct {
	DeviceType string
	ImageID    string
}
type optionalMetadata map[string]interface{}

func (m *Metadata) Write(p []byte) (n int, err error) {
	if err := decode(p, &m.Optional); err != nil {
		return 0, err
	}
	if err := decode(p, &m.Required); err != nil {
		return 0, err
	}
	return len(p), nil
}
```
I have tested that it works. The only drawback is that at the end we will have some data duplicated across optional and required fields, but the writing and reading is simple. To remove duplicates we will have to figure out some custom marshaling that will remove from optional what is in required.
But since most of the time (at least server) will take care only what is there in required part it should work as proposed.
Let me know what do you think.

pasinskim:
Since there are no more comments I am going to merge it.

PR #18, opened 2016-09-26
Parser: GetImageID method

kjaskiewiczz:
New rootfs parser method - GetImageID.
Method similar to GetDeviceType.

Required metadata fields are DeviceType and ImageID.
There is GetDeviceType() method already, but there was no GetImageID() method.

Signed-off-by: Krzysztof Jaskiewicz <krzysztof.jaskiewicz@open-rnd.pl>

### DIFF
--- a/parser/rootfs_parser.go
+++ b/parser/rootfs_parser.go
@@ -62,6 +62,9 @@ func (rp *RootfsParser) GetUpdateFiles() map[string]UpdateFile {
 func (rp *RootfsParser) GetDeviceType() string {
 	return rp.metadata.Required.DeviceType
 }
+func (rp *RootfsParser) GetImageID() string {
+	return rp.metadata.Required.ImageID
+}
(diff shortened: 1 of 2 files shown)


pasinskim:
Actually I am not sure if this is the right way of extending this library. What is some new parser won't support image ID at all? If you need this functionality, maybe better solution would be extending `rootfs-image` parser as we know we will need image ID there.

kjaskiewiczz:
I thought that both image id and device type are required and all parsers should support obtaining both, not device type only.
If image id is optional then why this field is in required metadata structure?

kacf:
Marcin and I discussed this. There is no completely obvious solution, this is all about anticipating what we might do with the format later. But we agreed that image ID probably should go inside the image-rootfs parser, and it's only a required field in the rootfs' metadata, not necessarily other update types.

Correct me if I'm wrong, @pasinskim.

kjaskiewiczz:
@pasinskim @kacf: updated

PR #20, opened 2016-09-27
Support for passing incoming update data stream to the caller

bboozzoo:
The user of `RootfsParser` can provide `DataFunc` to receive a decompressed update data stream along with information about device type and current update.

Additional minor enhacements that remove the need for specialized `New..Parser`.

Do not merge yet, still missing tests.

@pasinskim @kacf

### DIFF
--- a/parser/generic_parser.go
+++ b/parser/generic_parser.go
@@ -99,7 +99,26 @@ func (rp *GenericParser) ParseHeader(tr *tar.Reader, hdr *tar.Header, hPath stri
 	return nil
 }
 
+// Default data parser what writes the uncompressed data to `w`.
 func parseData(r io.Reader, w io.Writer, uFiles map[string]UpdateFile) error {
+	return parseDataWithHandler(
+		r,
+		func(dr io.Reader, uf UpdateFile) error {
+			if _, err := io.Copy(w, dr); err != nil {
+				return errors.Wrapf(err, "parser: can not read data: %v", uf.Name)
+			}
+			return nil
+		},
+		uFiles,
+	)
+}
+
+// Caller provided update data stream handler. Parameter `r` is a decompressed
+// data stream, `uf` contains basic information about update. The handler shall
+// return nil if no errors occur.
+type parseDataHandlerFunc func(r io.Reader, uf UpdateFile) error
+
+func parseDataWithHandler(r io.Reader, handler parseDataHandlerFunc, uFiles map[string]UpdateFile) error {
 	if r == nil {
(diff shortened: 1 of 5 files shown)


parser/rootfs_parser.go
```diff
@@ -30,26 +30,25 @@ import (
...
 
+// User provided update data stream handler. Parameter `r` is a decompressed
+// data stream, `dt` holds current device type, `uf` contains basic information
+// about update. The handler shall return nil if no errors occur.
+type DataHandlerFunc func(r io.Reader, dt string, uf UpdateFile) error
+
 type RootfsParser struct {
-	dataWriter io.Writer
-	scriptDir  string
+	W         io.Writer
```
pasinskim: Maybe we can have something more descriptive here since we are exporting it.

parser/generic_parser.go
```diff
@@ -101,7 +99,24 @@ func (rp *GenericParser) ParseHeader(tr *tar.Reader, hdr *tar.Header, hPath stri
...
+			}
+			return nil
+		},
+		uFiles)
+}
+
+// Caller provided update data stream handler. Parameter `r` is a decompressed
+// data stream, `uf` contains basic information about update. The handler shall
+// return nil if no errors occur.
+type parseDataHandlerFunc func(r io.Reader, uf UpdateFile) error
```
pasinskim: I think this should be exported. This will make it more obvious that you need to implement this yourself to use wit some custom parser type.

parser/rootfs_parser.go
```diff
@@ -30,26 +30,25 @@ import (
 	"github.com/pkg/errors"
 )
 
+// User provided update data stream handler. Parameter `r` is a decompressed
```
pasinskim: Start the comment with `DataHandlerFunc is user ...`

parser/generic_parser.go
bboozzoo: ok

parser/rootfs_parser.go
bboozzoo: ok

parser/rootfs_parser.go
bboozzoo: Trying to follow the naming pattern established by stdlib, for instance: https://godoc.org/io#LimitedReader

I'll add a comment that W is where data is written to.

parser/generic_parser.go
bboozzoo: Actually not ok. Neither `parseData` nor `parseDataWithHandler` are exported, so there's little sense in exporting a type that you cannot use.

PR #22, opened 2016-09-28
Add vendored dependencies.

pasinskim:


maciejmrowiec:
@pasinskim what tool did you use for vendoring ?

pasinskim:
@maciejmrowiec govendor.

kacf:
[This](https://github.com/mendersoftware/artifacts/pull/19) should be merged first, and you should retest this to get the licenses right. And you should use `godep` to add the dependencies, since this is what we have used so far.

PR #26, opened 2016-10-03
Add basic CLI for storing artifacts. 

pasinskim:


artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"path/filepath"
+
+	"github.com/mendersoftware/artifacts/parser"
+	"github.com/mendersoftware/artifacts/writer"
+
+	"github.com/urfave/cli"
+)
+
+func wrieArtifact(c *cli.Context) error {
```
bboozzoo: write

artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+	if c.NArg() != 3 ||
+		len(c.String("device-type")) == 0 || len(c.String("image-id")) == 0 {
+		return errors.New("invalid arguments")
+	}
+	aw := awriter.NewWriter("mender", 1)
+	p := parser.RootfsParser{}
+	aw.Register(&p)
+
+	updateTestDir, _ := ioutil.TempDir("", "update")
+	ioutil.WriteFile(filepath.Join(updateTestDir, "meta-data"), []byte(`{"deviceType": "`+c.Args().Get(1)+`", "imageId": "`+c.Args().Get(2)+`"}`), os.ModePerm)
```
bboozzoo: IIRC `c.Args().Get(1)` is not the same as `c.String("device-type")`

artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return aw.WriteSingle(updateTestDir, c.Args().First(), "rootfs-image", filepath.Join(updateTestDir, "mender.tar.gz"))
```
bboozzoo: Can you break the lines at 80 chars? It's hard to view the diff on github.

artifacts.go
```diff
@@ -0,0 +1,106 @@
...
+
+	if c.NArg() != 3 ||
+		len(c.String("device-type")) == 0 || len(c.String("image-id")) == 0 {
+		return errors.New("invalid arguments")
+	}
+	aw := awriter.NewWriter("mender", 1)
+	p := parser.RootfsParser{}
+	aw.Register(&p)
+
+	updateTestDir, _ := ioutil.TempDir("", "update")
```
bboozzoo: It would be nice if the user could pass the location of output directory

artifacts.go
bboozzoo: Maybe print the output file path to the to stdout? It's easier to use it in the script if the tools tells you where it stored the file. Otherwise the caller will have to make an assumption that about the name and location of the output artifact.

artifacts.go
pasinskim: fixed!

artifacts.go
pasinskim: This was done in some of the subsequent commits.

artifacts.go
pasinskim: Superseded by new commit.

artifacts.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func validateArtifact(c *cli.Context) error {
+	if c.NArg() == 0 {
+		return errors.New("Nothing specified, nothing validated. \nMaybe you wanted" +
+			"to say 'artifacts validate <pathspec>'?")
+	}
+	_, err := os.Stat(c.Args().First())
+	if err != nil {
+		return errors.New("Pathsec '" + c.Args().First() +
```
bboozzoo: pathspec?

artifacts.go
```diff
@@ -0,0 +1,169 @@
...
+		cli.StringFlag{
+			Name:  "device-type, t",
+			Usage: "Type of device supported by the update",
+		},
+		cli.StringFlag{
+			Name:  "image-id, i",
+			Usage: "Yocto id of the update image",
+		},
+		cli.StringFlag{
+			Name:  "name, n",
```
bboozzoo: I'm wondering if we should replace `--name|-n` with `--output-path|-o`. Output seems more appropriate. Name, on the other hand, is bit confusing. One does not know whether is a path or something else, or where the output will be written to.

artifacts.go
```diff
@@ -0,0 +1,169 @@
...
+	if len(c.String("name")) > 0 {
+		name = c.String("name")
+	}
+
+	aw := awriter.NewWriter("mender", 1)
+	return aw.WriteKnown([]parser.UpdateData{ud}, name)
+}
+
+func readArtifact(c *cli.Context) error {
+	return nil
```
bboozzoo: maybe return `errors.New("not implemented")`

parser/parser_manager.go
```diff
@@ -32,6 +32,19 @@ type UpdateFile struct {
 	Signature []byte
 }
 
+type UpdateData struct {
+	// path to directory containg update data
+	Path string
+	// update data files
```
bboozzoo: paths to update data files

artifacts.go
pasinskim: I have stolen it form git. It is path specifier. Can change to something else.

artifacts.go
pasinskim: Fixing!

artifacts.go
bboozzoo: i meant there is a typo, `Pathsec`, I guess it was supposed to be `Pathspec`?

artifacts.go
bboozzoo: Yup, will do

parser/parser_manager.go
pasinskim: Or not. This is actually a path to the directory that might contain not only data files. There might be `meta-data`, `type-info` files and so on. I think it will be a bit misleading to say that this is where data files are.

bboozzoo: Sounds like you need to update the comment anyway.

pasinskim: Will do.

PR #35, opened 2016-11-21
MEN-809 Add checksum file extension.

pasinskim:
This is fixing the missing original data file extension (e.g. ext4)
before adding the checksum file extension.
In the former case for the update.ext4 file we had update.sha256sum
checksum file. After the fix the checksum file is update.ext4.sha256sum.


kacf:
Nice! I'll try to run this with the test I wrote.

PR #36, opened 2016-11-28
Fix reading artifact containing multiple updates.

pasinskim:


kjaskiewiczz:
I don't understand how codecov works.
Artifacts had ~73% coverage before this patch.
Diff is covered in 60.00% and but the total coverage was increased by 0,24%

PR #37, opened 2016-12-01
Use the exact update type as a type-info parameter.

pasinskim:
So far we've been returning the parser type as a update-type for a given update. This was correct for the specific update type parsing (like `rootfs-image`), as to use a given parser we need to have a match between parser parsing type and update type itself.
For the generic parser though, we've been returning `generic` as a given update type even though the exact type was different.
There are also some fixes and improvements added:
- missing space in error messages while using CLI tool
- always using `GenericParser` as a default parser type

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -17,7 +17,6 @@ package main
 import (
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


kacf:
Is the `typeParsed` variable actually used in the logic anywhere, or is there nothing rootfs specific to the parsing?

The only check I can think of which would be rootfs specific, is that rootfs updates should enforce one update bucket only. But I'm not sure if we make this check right now.

pasinskim:
@kacf not sure if I got it. 
`typeParsed` is used *ONLY* when we are using `GenericParser`. This one, you are using if you are not having a parser for the specific type you are parsing. 
If we are using `RootfsParser` from the other hand, the type parsed *MUST BE* the same as a parser type, thus `typeParsed` is not needed.

And the check you are mentioning is already there. `RootfsParser` allows only one update file.

kacf:
Ah, yes, I see now. I thought `GenericParser` was a part of of `RootfsParser`, and hence would be used by it, but I see now that it's not. Never mind then!

PR #38, opened 2016-12-01
travis: upload binary to s3 bucket

GregorioDiStefano:
@eysteins was trying to get this working locally.

I think its a good idea to just deploy this to s3 whenever tests are passing.

@pasinskim what do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
     # Install code coverage / coveralls tooling
     - go get -u github.com/axw/gocov/gocov
     - go get -u golang.org/x/tools/cmd/cover
-
+    - go get ./...
 before_script:
     # Print build info.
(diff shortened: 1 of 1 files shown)


pasinskim:
Maybe, yes. Makes sense as we shouldn't require users to compile this themselves. Probably we should also add some docs on how to download and use it.

GregorioDiStefano:
Yes, precisely. @eysteins is writing docs on how to compile this, but we can just provide a build from the build machine instead.

pasinskim:
@GregorioDiStefano but we will need to change the configuration to actually build the binary. Previously it was supposed to be a library so we are not providing an actual `artifacts` binary AFAIK.

GregorioDiStefano:
updated so artifacts tool is compiled.

PR #40, opened 2016-12-16
Rename artifacts to mender-artifact and simplify CLI.

pasinskim:

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -18,7 +18,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 
 	"github.com/mendersoftware/artifacts/parser"
 	"github.com/mendersoftware/artifacts/reader"
@@ -26,6 +25,9 @@ import (
 
 	"github.com/urfave/cli"
 )
+
+// VERSION of the mender-artifact CLI tool
+var VERSION = "0.1"
 
 func writeArtifact(c *cli.Context) error {
 	if len(c.String("device-type")) == 0 || len(c.String("artifact-name")) == 0 ||
@@ -49,9 +51,9 @@ func writeArtifact(c *cli.Context) error {
 		name = c.String("output-path")
 	}
 
-	devices := strings.Split(c.String("device-type"), ",")
+	devices := c.StringSlice("device-type")
 
-	aw := awriter.NewWriter("mender", 1, devices, c.String("artifact-name"))
+	aw := awriter.NewWriter("mender", c.Int("version"), devices, c.String("artifact-name"))
 	return aw.WriteKnown([]parser.UpdateData{ud}, name)
 }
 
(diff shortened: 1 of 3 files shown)


artifacts.go
```diff
@@ -27,6 +26,9 @@ import (
 	"github.com/urfave/cli"
 )
 
+// VERSION of the mender-artifact CLI tool
+var VERSION = "0.1"
```
kacf: Maybe we should consider using the [same mechanism that we do for Mender](https://github.com/mendersoftware/mender/blob/master/Makefile#L18-L21). Hardcoded strings have a tendency to go out of date. What do you think?

pasinskim: Agree, but this is a bit tricky. Maybe instead of sha being a part of the version we can use branch? But for now this will be `master` which is not optimal either.

pasinskim: @kacf this is quite isolated and can be handled separately: https://tracker.mender.io/browse/MEN-897

kacf: Sure.

artifacts.go
```diff
@@ -150,52 +155,61 @@ func run() error {
...
 		},
 		cli.StringFlag{
 			Name:  "output-path, o",
-			Usage: "Full path to output artifact file",
+			Usage: "Full path to output artifact file.",
+		},
+		cli.IntFlag{
+			Name:  "version, v",
+			Usage: "Version of the artifact.",
+			Value: 1,
```
kacf: I don't think this will be configurable, will it? Or are you anticipating one tool being able to write several distinct versions?

artifacts.go
```diff
@@ -150,52 +155,61 @@ func run() error {
...
+	//
+	// validate
+	//
 	validate := cli.Command{
 		Name:        "validate",
-		Usage:       "Validates artifact file",
+		Usage:       "Validates artifact file.",
 		Action:      validateArtifact,
-		UsageText:   "atrifacts validate [options] <pathspec>",
+		UsageText:   "mender-atrifact validate [options] <pathspec>",
```
kacf: `s/atrifact/artifact/`

artifacts.go
```diff
@@ -150,52 +155,61 @@ func run() error {
...
+	}
+
+	//
+	// read
+	//
+	read := cli.Command{
+		Name:        "read",
+		Usage:       "Reads artifact file.",
+		Action:      readArtifact,
+		UsageText:   "mender-atrifact read [options] <pathspec>",
```
kacf: `s/atrifact/artifact/`

artifacts.go
pasinskim: Yes. The later. For now we are using only version 1, but I think soon it might be different.

artifacts.go
pasinskim: Thanks. Fixed!

artifacts.go
pasinskim: Fixed!

PR #45, opened 2017-01-16
MEN-952 Fix reader panic while reading malformed tar archive.

pasinskim:
After tar reader returns error, `hdr` variable might be left unset
in `ReadNextDataFile` and dereferencing it to log the `hdr.Name`
ends up with panic.


kacf:
We should turn off the codecov/patch metric, it's completely useless (code which happens not to be covered by a unit test cannot be changed!?)

PR #47, opened 2017-01-20
Reduce cyclomatic complexity.

pasinskim:
After adding cyclomatic complexity check it turned out that some functions
were having too much returns. This is fixing the issue making
`gocyclo -over 15` to pass.


kacf:
Included in #46.

PR #54, opened 2017-02-15
MEN-1028 don't crash when device type in not provided.

pasinskim:
While using `mender-artifact` CLI it crashed without providing
device type.


kacf:
Should be cherry-picked as well.

PR #55, opened 2017-02-15
Update 1.0.x

pasinskim:


pasinskim:
There are some issues with merging to 1.0.x instead of master. I will merge and test the result afterwards.

PR #60, opened 2017-03-14
Add basic artifact signig. 

pasinskim:


Documentation/artifact-format.md
```diff
@@ -95,6 +89,38 @@ Currently there is only version 1, but this document may describe later versions
...
 
+manifest
+----
+
+Format: text
+
+Contains the below content exactly:
+
+```
+Version: 2
```
kacf: Where does this version come from?

Documentation/artifact-format.md
```diff
@@ -95,6 +89,38 @@ Currently there is only version 1, but this document may describe later versions
...
+Date: 2017-03-13 20:15:40.800231486 +0100 CET
+SHA256:
+ 1d0b820130ae028ce8a79b7e217fe505a765ac394718e795d454941487c53d32 0000/update.ext4
+ 4d480539cdb23a4aee6330ff80673a5af92b7793eb1c57c4694532f96383b619 header.tar.gz
+ 52c76ab66947278a897c2a6df8b4d77badfa343fec7ba3b2983c2ecbbb041a35 version
```
kacf: Just out of curiosity, was this generated by some tool? I'm tempted to say we should stick to a standard format, which would mean to remove `SHA256:` and everything above it, remove the leading space, and have an extra space between the checksum and the file, like this:

```
1d0b820130ae028ce8a79b7e217fe505a765ac394718e795d454941487c53d32  0000/update.ext4
4d480539cdb23a4aee6330ff80673a5af92b7793eb1c57c4694532f96383b619  header.tar.gz
52c76ab66947278a897c2a6df8b4d77badfa343fec7ba3b2983c2ecbbb041a35  version
```

This matches the output of the `sha256sum` tool, which would be quite convenient. It's what we use for our license checking.

Documentation/artifact-format.md
```diff
@@ -95,6 +89,38 @@ Currently there is only version 1, but this document may describe later versions
...
+The manifest file contains checksums of compressed header, version and all
+data files being a part of the artifact.
+It contains also version information which is the same as the artifact
+version and date and time of creation of the file.
+
+
+manifest.sig
+----
+
+Format: TBD
```
kacf: I guess this format will be OpenSSL's signature format?

Documentation/artifact-format.md
pasinskim: I've been thinking that we will use the same as artifact version, but probably it is redundant here. We are already having a version in `version` file.

Documentation/artifact-format.md
pasinskim: See this: https://tracker.mender.io/browse/MEN-1016?focusedCommentId=78528&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-78528.

Maybe we can do only checksums here in more "standard" way. I am not using any library for this, so we can decide how this should look like.

Documentation/artifact-format.md
pasinskim: If we will decide that we are using OpenSSL. Once this will be closed https://tracker.mender.io/browse/MEN-1110 I will update the docs.

Documentation/artifact-format.md
kacf: Ok, then I suggest the format above. It's not a strong requirement, mostly convenience, and maybe it will be a benefit later.

pasinskim: Makes a lot of sense. Fixed!

Documentation/artifact-format.md
pasinskim: Fixed! Version and all things but checksums removed.

kacf:
Superseded by #61.

PR #65, opened 2017-04-28
Add a state scripts support.

pasinskim:


awriter/writer.go
```diff
@@ -211,6 +228,13 @@ func writeHeader(tw *tar.Writer, devices []string, name string,
 		return errors.New("writer: can not store header-info")
 	}
 
+	// write scripts
+	if scr != nil {
+		if err := writeScripts(tw, scr); err != nil {
+			return err
+		}
+	}
+
```
kacf: Scripts are not valid in artifact v1. Does this need to be version protected?

artifacts.go
```diff
@@ -32,6 +32,39 @@ import (
 // Version of the mender-artifact CLI tool
 var Version = "unknown"
 
+func version(c *cli.Context) int {
+	version := c.Int("version")
+	// set version to 2 if artifact is signed and version
+	// is not set explicitly
+	if !c.IsSet("version") && len(c.String("key")) != 0 {
+		version = 2
+	}
```
kacf: You need to do this for scripts too, right?

artifacts.go
```diff
@@ -32,6 +32,39 @@ import (
...
+		}
+		return awriter.NewWriterSigned(f, artifact.NewSigner(privateKey)), nil
+	}
+	return awriter.NewWriter(f), nil
+}
+
+func scripts(c *cli.Context) (*artifact.Scripts, error) {
+	scr := artifact.Scripts{}
+	if len(c.StringSlice("script")) > 0 {
+		for _, script := range c.StringSlice("script") {
```
kacf: You might be able to lose the if here. If the range is zero it should just skip over the loop.

artifact/scripter.go
```diff
@@ -0,0 +1,58 @@
...
+
+var availableScriptType = map[string]bool{
+	"Idle":                true,
+	"Sync":                true,
+	"ArtifactDownload":    true,
+	"ArtifactPreinstall":  true,
+	"ArtifactInstall":     true,
+	"Reboot":              true,
+	"ArtifactPostinstall": true,
+	"ArtifactCommit":      true,
```
pasinskim: Thanks for spotting this. Probably I took a look at some subset `Order that states will usually be executed in:`. Anyway, fixed now and missing states are added.

awriter/writer.go
pasinskim: I don't think this needs to be protected here. The better place would be in `artifacts.go` where you are writing artifact reading command line parameters. I will add a check there.

artifacts.go
pasinskim: Yup. Will add a check for this.

artifacts.go
pasinskim: Fixing!

pasinskim:
After rerunning build was green. The first failed probably because some timing issues.

PR #66, opened 2017-05-04
MEN-1198 Stop reading artifact when signature is missing.

pasinskim:
If we are expecting artifact to be signed, `NewReaderSigned`
instead of `NewReader` should be used. This will allow failing
immediately after discovering that signature is missing.

Changelog: None

### DIFF
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -158,6 +158,10 @@ func (ar *Reader) GetHandlers() map[int]handlers.Installer {
 }
 
 func (ar *Reader) readHeaderV1(tReader *tar.Reader) error {
+	if ar.signed {
+		return errors.New("reader: expecting signed artifact; " +
+			"v1 is not supporting signatures")
+	}
 	hdr, err := getNext(tReader)
 	if err != nil {
 		return errors.New("reader: error reading header")
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -55,6 +55,15 @@ func NewReader(r io.Reader) *Reader {
 	}
 }
 
+func NewReaderSigned(r io.Reader) *Reader {
+	return &Reader{
+		r:          r,
+		signed:     true,
+		handlers:   make(map[string]handlers.Installer, 1),
+		installers: make(map[int]handlers.Installer, 1),
+	}
+}
+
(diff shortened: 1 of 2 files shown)


kacf:
@pasinskim: Actually, it doesn't seem like this helped. It's still writing the update. I updated mender-artifact and added the same changes into the mender vendor directory. That should be enough, no?

kacf:
Hmm, wait I think I made a mistake, I will try again.

kacf:
Ok, so your fix, together with https://github.com/mendersoftware/mender/pull/205, fixed the issue the bug was about, but a different case has regressed. Check this output:

```
E                   Failed: Update succeeded when it should not have: Not signed, key present, Output: time="2017-05-05T09:03:54Z" level=info msg="device key path not configured, fallback to default mender-agent.pem" module=config 
E                   time="2017-05-05T09:03:54Z" level=info msg="Start updating from local image file: [image.mender]" module=rootfs 
E                   Installing update from the artifact of size 4096
E                   time="2017-05-05T09:03:54Z" level=info msg="opening device /dev/mmcblk0p3 for writing" module="block_device" 
E                   time="2017-05-05T09:03:54Z" level=info msg="partition /dev/mmcblk0p3 size: 109051904" module="block_device" 
E                   time="2017-05-05T09:03:54Z" level=info msg="wrote 23/23 bytes of update to device /dev/mmcblk0p3" module=device 
E                   .                                100% 4 KiB
E                   time="2017-05-05T09:03:54Z" level=info msg="Enabling partition with new image installed to be a boot candidate: 3" module=device
```

So now it allows an update that is not signed, even if the key is present.

I compared this against a run where I removed your changes, and there the test passes.

pasinskim:
@kacf are you using v1 artifact in test? This is the only way I can see this to happen.

kacf:
Passing locally, running all three PRs together [here](https://ci.cfengine.com/job/yoctobuild-pr/527/):
* https://github.com/mendersoftware/mender-artifact/pull/66
* https://github.com/mendersoftware/mender/pull/205
* https://github.com/mendersoftware/meta-mender/pull/291

pasinskim:
Retested and passed: https://ci.cfengine.com/job/yoctobuild-pr/529/

PR #71, opened 2017-05-30
Update state scripts to new naming schema.

pasinskim:
Use `STATE_[Enter|Leave|Error]_NUM[_optional-script-name]` schema
for naming state scripts.

Changelog: None


pasinskim:
@kacf maybe not extremely important, but maybe we should cherry-pick to 2.0.x as well?

kacf:
@pasinskim: Yes, I think we should. I'll leave that to you though. :-)

PR #80, opened 2017-06-19
MEN-1220 Sign existing artifacts using mender-artifact CLI

pasinskim:
Changelog: Title

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -21,13 +21,14 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/mendersoftware/mender-artifact/areader"
 	"github.com/mendersoftware/mender-artifact/artifact"
 	"github.com/mendersoftware/mender-artifact/awriter"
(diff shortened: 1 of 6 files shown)


artifacts.go
```diff
@@ -311,14 +312,203 @@ func validateArtifact(c *cli.Context) error {
...
+		if getNext {
+			raw, err = r.ReadRaw()
+			if err != nil {
+				return err
+			}
+		}
+		getNext = true
+
+		// check if we are not having element out of order
+		if elem.required && !strings.HasPrefix(raw.Name, elem.name) {
```
pasinskim: This is basically what we are doing in all other places. This is in case we will change the compression algorithm from gz to something else.

kacf: Oh, good thinking! I would change it to `header.tar.` then, but otherwise leave it as is. I don't think the tar format will change.

pasinskim:
After updating tests everything is working fine: https://mender-jenkins.mender.io/job/mender-builder/1357/

PR #83, opened 2017-07-25
MEN-1328: Mender-Artifact now returns an error code to the os on cli …

oleorhagen:
…errors

Changelog: Title

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -637,5 +637,7 @@ func run() error {
 }
 
 func main() {
-	run()
+	if err := run(); err != nil {
+		os.Exit(1)
+	}
 }


drewmoseley:
I was able to verify this by hand.  It will need to be cherry-picked back to the 1.0 and 2.0 branches to really fix the issue I was seeing when using the Yocto morty branch.

olepor:
@drewmoseley I'm pretty sure state scripts are only supported for 2.0 :)

drewmoseley:
Understood but the lack of an error return code would cause a problem with _any_ unrecognized options.  In my case I just happened to notice by trying to enable the example-state-scripts recipe on a version that didn't have support for them.

oleorhagen:
Fair point 😏

artifacts.go
```diff
@@ -637,5 +637,7 @@ func run() error {
 }
 
 func main() {
-	run()
+	if err := run(); err != nil {
+		os.Exit(-1)
```
kacf: Should return a positive return code, negative return codes are not supported by POSIX AFAIK, and their use is discouraged. Negative numbers may be used to communicate other aspects, such as whether the process died from a signal, so we should stay away from that range.

oleorhagen: What return value would you like?

kacf: Just `1` is okay. :-)

kacf:
Hmm, why does the license check fail now...? It's up to date as far as I can see.

PR #86, opened 2017-08-21
MEN-1355: mender-artifact now fails with whitespace in the artifact-name

oleorhagen:
Changelog: Title

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -88,11 +88,6 @@ func scripts(c *cli.Context) (*artifact.Scripts, error) {
 }
 
 func writeArtifact(c *cli.Context) error {
-	if len(c.StringSlice("device-type")) == 0 ||
-		len(c.String("artifact-name")) == 0 ||
-		len(c.String("update")) == 0 {
-		return cli.NewExitError("must provide `device-type`, `artifact-name` and `update`", 1)
-	}
 
 	// set default name
 	name := "artifact.mender"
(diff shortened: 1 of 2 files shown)


kacf:
Alright, QEMU test passed. It looks like Travis is failing because it measures the [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) is too high. A bit unfair in this case, since your change is tiny. Probably the function was very close to the limit already. Maybe you can get around this by adding the check to the parameter validation instead of during writing?

artifacts.go
```diff
@@ -523,8 +523,13 @@ func run() error {
...
 	writeRootfs := cli.Command{
-		Name:   "rootfs-image",
-		Action: writeArtifact,
+		Name: "rootfs-image",
+		Action: func(c *cli.Context) error {
+			if len(strings.Fields(c.String("artifact-name"))) > 1 { // check for whitespace in artifact-name
+				return cli.NewExitError("whitespace is not allowed in the artifact-name", 1)
+			}
+			return writeArtifact(c)
+		},
```
oleorhagen: @kacf took a slightly different approach when rewriting this, and decided to wrap the writeartifact function. Not super clean, but acceptable?

kacf: No, I think this is quite clean actually, as long as it's as small as that!

kacf:
@oleorhagen, @GregorioDiStefano: This failed twice with seemingly the same error:

* https://mender-jenkins.mender.io/job/mender-builder/2108/console
* https://mender-jenkins.mender.io/job/mender-builder/2114/console

Greg, the tests are otherwise in a good state, right? Could it be a real error in the PR?

PR #95, opened 2017-09-06
MEN-1412 Fix broken header checksum verification.

pasinskim:
For reading and verifying header checksum we are using checksum.Reader.
Internal inplementation of checksum.Reader is using io.TeeReader for
calculating checksum and reading the stream at the same time.
After receiving EOF checksum should be checked automatically inside
checksum.Reader, but if we are inside tar archive error is not passed up
the stack to tar.Reader.
The workaround is to check the sum again after the whole header is read.

Changelog: Title

### DIFF
--- a/artifact/checksum.go
+++ b/artifact/checksum.go
@@ -77,12 +77,8 @@ func (c *Checksum) Read(p []byte) (int, error) {
 	n, err := c.r.Read(p)
 	if err == io.EOF {
 		// verify checksum
-		sum := c.h.Sum(nil)
-		checksum := make([]byte, hex.EncodedLen(len(sum)))
-		hex.Encode(checksum, sum)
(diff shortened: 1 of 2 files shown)


artifact/checksum.go
```diff
@@ -98,6 +94,15 @@ func (c *Checksum) Checksum() []byte {
 	return checksum
 }
 
+func (c *Checksum) Verify() error {
+	sum := c.Checksum()
```
kacf: Does this calculate the checksum for the whole file or it just grabs the running sum which has currently accumulated for the stream?

pasinskim: The later. That's why this is important to call this function when we see EOF for any given part of the stream.

pasinskim:
@kacf PR updated. Now it's passing tests.

PR #97, opened 2017-11-15
MEN-1213 

pasinskim:
Modify existing images.


debugfs.go
```diff
@@ -0,0 +1,73 @@
...
+
+	return tmpDir, nil
+}
+
+func debugfsReplaceFile(imageFile, newFile, image string) error {
+	scr, err := ioutil.TempFile("", "mender-debugfs")
+	if err != nil {
+		return errors.Wrap(err, "debugfs: create sync script file")
+	}
+	defer os.Remove(scr.Name())
```
kacf: Needs `scr.Close()` in all return paths prior to natural `scr.Close()`. See my other more general comment (search for "Go drawback").

artifacts.go
```diff
@@ -301,205 +294,477 @@ func validateArtifact(c *cli.Context) error {
...
-			continue
+	data := dataFile
+	ar := areader.NewReader(from)
+
+	if dataFile == "" {
+		tmpData, tmpErr := ioutil.TempFile("", "mender-repack")
+		if tmpErr != nil {
+			return nil, tmpErr
 		}
+		defer os.Remove(tmpData.Name())
```
kacf: Needs `defer tmpData.Close()` after `Remove` (which confusingly will run it first).

artifacts.go
```diff
@@ -301,205 +294,477 @@ func validateArtifact(c *cli.Context) error {
...
+}
+
+// oblivious to whether the file exists beforehand
+func modifyName(name, image string) error {
+	data := fmt.Sprintf("artifact_name=%s", name)
+	tmpNameFile, err := ioutil.TempFile("", "mender-name")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpNameFile.Name())
```
kacf: Needs `tmpNameFile.Close()` in all return paths prior to natural `tmpNameFile.Close()`. See my other more general comment (search for "Go drawback").

artifacts.go
```diff
@@ -301,205 +294,477 @@ func validateArtifact(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	defer art.Close()
+
+	tmp, err := ioutil.TempFile("", "mender-artifact")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmp.Name())
```
kacf: Needs `defer tmp.Close()` after `Remove`.

artifacts.go
```diff
@@ -301,205 +294,486 @@ func validateArtifact(c *cli.Context) error {
...
-	case 2:
-		tFile, err := ioutil.TempFile("", "mender-artifact")
-		if err != nil {
-			return "", errors.Wrap(err,
-				"Can not create temporary file for storing artifact")
+	reg := regexp.MustCompile(`(?m)^[[:blank:]][0-9]+[[:blank:]]+([0-9]+)s[[:blank:]]+[0-9]+s[[:blank:]]+([0-9]+)s`)
+	partitionMatch := reg.FindAllStringSubmatch(string(out), -1)
+
+	// IMPORTANT: we are assuming standard Mender formating here:
+	// only 2nd and 3rd partitions are rootfs we are going to modify
```
kacf: Not sure how much work it is, but there is a simple test/assertion you can do: Just verify that `/etc/mender/artifact_info` exists on the partition.

artifacts.go
```diff
@@ -301,205 +294,486 @@ func validateArtifact(c *cli.Context) error {
...
+		return partitions, nil
+		// if we have single ext file there is no need to mount it
+	} else if len(partitionMatch) == 1 {
+		return []partition{{path: image}}, nil
+	}
+	return nil, fmt.Errorf("invalid partition table: %s", string(out))
+}
 
-		err = signArtifact(aReader, aWriter, data, key, force)
+func mountSdimg(partitions []partition, image string) error {
```
kacf: Suggest `extractFromSdimg` as function name instead. Mounting is not really what it does.

modify_exisintg_test.go
```diff
@@ -0,0 +1,216 @@
...
+	if err != nil {
+		return false
+	}
+	defer os.RemoveAll(tmp)
+
+	data, err := ioutil.ReadFile(filepath.Join(tmp, filepath.Base(file)))
+	if err != nil {
+		return false
+	}
+	fmt.Printf("expected: [%s]; actual: [%s]\n", expected, string(data))
```
kacf: Leftover debug?

areader/reader.go
```diff
@@ -217,7 +217,6 @@ func readManifest(tReader *tar.Reader) (*artifact.ChecksumStore, error) {
 
 func signatureReadAndVerify(tReader *tar.Reader, message []byte,
 	verify SignatureVerifyFn, signed bool) error {
-
```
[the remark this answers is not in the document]
pasinskim: It's how it is... 1.9 seems to be even more strict with those things.

artifacts.go
```diff
@@ -301,207 +292,194 @@ func validateArtifact(c *cli.Context) error {
...
-	rawVer *artifact.Raw, key []byte, force bool) error {
-	// first we need to store version in new artifact we are trying to sign
-	if err := w.WriteRaw(rawVer); err != nil {
-		return err
+	tFile, err := ioutil.TempFile("", "mender-artifact")
+	if err != nil {
+		return errors.Wrap(err,
+			"Can not create temporary file for storing artifact")
 	}
+	defer os.Remove(tFile.Name())
```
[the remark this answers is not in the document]
pasinskim: Actually I can do defer. It will simply fail if the file is already closed, but I don't care at that point. The important thing is we are not leaking resources.

artifacts.go
pasinskim: Fixed!

artifacts.go
pasinskim: Fixed!

artifacts.go
pasinskim: Fixed!

artifacts.go
pasinskim: It is done elsewhere, where we are actually modifying something. Here we would need to dd the partitions to some temporary place and check if the file exists. Might be quite heavy especially because it is verified later on anyway.

artifacts.go
pasinskim: Makes sense. Renamed!

debugfs.go
pasinskim: Fixed!

modify_exisintg_test.go
```diff
@@ -0,0 +1,113 @@
+// Copyright 2017 Northern.tech AS
```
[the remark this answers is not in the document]
pasinskim: Fixed!

modify_exisintg_test.go
pasinskim: Yup. Removed.

pasinskim:
@kacf all should be fixed now.

PR #100, opened 2017-11-30
use cwd instead of os's temp dir for creating artifacts

GregorioDiStefano:
Signed-off-by: Gregorio Di Stefano <greg.distefano+github@gmail.com>

### DIFF
--- a/artifacts.go
+++ b/artifacts.go
@@ -363,7 +363,13 @@ func signExisting(c *cli.Context) error {
 		return cli.NewExitError("Can not use signing key provided: "+err.Error(), 1)
 	}
 
-	tFile, err := ioutil.TempFile("", "mender-artifact")
+	cwd, err := os.Getwd()
+
+	if err != nil {
+		return err
+	}
+
+	tFile, err := ioutil.TempFile(cwd, "mender-artifact")
 	if err != nil {
 		return errors.Wrap(err,
 			"Can not create temporary file for storing artifact")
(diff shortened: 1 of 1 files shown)


artifacts.go
```diff
@@ -791,7 +818,7 @@ func getCandidatesForModify(path string, key []byte) ([]partition, bool, error)
 func modifyArtifact(c *cli.Context) error {
 	if c.NArg() == 0 {
 		return cli.NewExitError("Nothing specified, nothing will be modified. \n"+
-			"Maybe you wanted to say 'artifacts read <pathspec>'?", 1)
+			"Maybe you wanted to say 'oartifacts read <pathspec>'?", 1)
```
kacf: Did you *really* want to say "oartifacts"? :-)

artifacts.go
```diff
@@ -394,10 +400,6 @@ func signExisting(c *cli.Context) error {
 		return cli.NewExitError("Unsupported version of artifact file: "+string(ver), 1)
 	}
 
-	if err = tFile.Close(); err != nil {
```
GregorioDiStefano: Do we need this? The closing of the fh is already deferred AND it reduces the cyclomatic complexity to 15 :)

kacf: Yes, we need it because the file is later renamed, and this happens before the deferred call.

GregorioDiStefano:
@

PR #101, opened 2017-12-11
Create temporary artifact in the directory where original one is

pasinskim:
If /tmp directory is not the same partition as the directory
where modified artifact is stored, we are ending up with the following
error while modifying artifact: invalid cross-device link.
This is storing temporary artifact file in the same location where
the original one is placed.

Changelog: None

Signed-off-by: Marcin Pasinski marcin.pasinski@northern.tech


GregorioDiStefano:
Yes, this works!

kacf:
Don't merge yet, this is not running in Jenkins for some reason...

PR #104, opened 2018-01-18
travis-ci: only upload mender-artifact binary

GregorioDiStefano:
Signed-off-by: Gregorio Di Stefano <greg.distefano@gmail.com>


GregorioDiStefano:
@maciejmrowiec we were uploading the entire workdir, when we only need the binary.

PR #107, opened 2018-01-29
MEN-1667 Artifact version 3 format documentation

pasinskim:
Add a new documentation for version 3 of Mender artifact.
This contains `artifact_provides` and `artifact_depends` sections
needed for updates ordering as well as some preliminary additions
for delta updates.

Changelog: Title

Signed-off-by: Marcin Pasinski <marcin.pasinski@northern.tech>

### DIFF
--- a/Documentation/artifact-format.md
+++ b/Documentation/artifact-format.md
@@ -8,6 +8,82 @@ Note that there are some restrictions on ordering of the files, described
 in the "Ordering" section.
 
 
+### version 3
+
+```
+-artifact.mender (tar format)
+  |
+  +---version
+  |
+  +---manifest
+  |
+  +---manifest.sig
+  |
+  +---manifest-augment
+  |
+  +---header.tar.gz (tar format)
+  |    |
+  |    +---header-info
+  |    |
+  |    +---scripts
+  |    |    |
+  |    |    +---State.Enter
+  |    |    +---State.Leave
+  |    |    +---State.Error
+  |    |    `---<more scripts>
+  |    |
+  |    `---headers
+  |         |
+  |         +---0000
+  |         |    |
+  |         |    +---files
+  |         |    |
+  |         |    +---type-info
+  |         |    |
+  |         |    +---meta-data
+  |         |
+  |         +---0001
+  |         |    |
+  |         |    `---<more headers>
+  |         |
+  |         `---000n ...
+  |
+  +---header-augment.tar.gz (tar format)
+  |    |
+  |    +---header-info
+  |    |
+  |    `---headers
+  |         |
+  |         +---0000
+  |         |    |
+  |         |    +---files
+  |         |    |
+  |         |    +---type-info
+  |         |    |
+  |         |    +---meta-data
+  |         |
+  |         +---0001
+  |         |    |
+  |         |    `---<more headers>
+  |         |
+  |         `---000n ...
+  |
+  `---data
+       |
+       +---0000.tar.gz
+       |    +--<image-file (ext4)>
+       |    +--<binary delta, etc>
+       |    `--...
+       |
+       +---0001.tar.gz
+       |    +--<image-file (ext4)>
+       |    +--<binary delta, etc>
+       |    `--...
+       |
+       +---000n.tar.gz ...
+            `--...
+```
+
 ### version 2
 
 ```
@@ -122,7 +198,7 @@ Contains the below content exactly:
 ```
 {
   "format": "mender",
-  "version": 2
+  "version": 3
 }
 ```
 
@@ -137,7 +213,7 @@ manifest
 Format: text
 Version: Exists only in version 2 and later
 
-Contains the file checksums, formatted exactly like below:
+Contains file checksums, formatted exactly like below:
 
 ```
 1d0b820130ae028ce8a79b7e217fe505a765ac394718e795d454941487c53d32  data/0000/update.ext4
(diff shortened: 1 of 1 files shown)


Documentation/artifact-format.md
```diff
@@ -154,14 +230,36 @@ the two spaces.
...
 It is legal for an artifact not to have signature file.
 
 
+manifest-augment
+----
+
+Format: text
+Version: Exists only in version 3 and later
+
+Contains the file checksums, formatted exactly like below:
```
kacf: "Contains ~~the~~ file checksums". Otherwise it sounds like it contains all checksums.

Documentation/artifact-format.md
```diff
@@ -8,6 +8,82 @@ Note that there are some restrictions on ordering of the files, described
...
+```
+-artifact.mender (tar format)
+  |
+  +---version
+  |
+  +---manifest
+  |
+  +---manifest.sig
+  |
+  +---manifest-augment
```
kacf: Now that I think of it, what do you think of renaming `manifest-augment` to `manifest-unsigned`? It might make it more obvious what it is for. And same for the header.

Documentation/artifact-format.md
```diff
@@ -175,8 +273,89 @@ version is specified outside of `header.tar.gz`.
...
+
+* `artifact_name` is the name of the artifact currently installed on the device
+* `rootfs_image_checksum` is the checksum of the image that needs to be installed
+on the device before current artifact can be installed
+* `device_type` is the type of the device (see `device_provides` below)
+* `artifact_group` is the group of the artifacts current artifact belongs to
+* `update_types_supported` is the list of the update types Mender agent installed
+on the device can handle
+* `artifact_versions_supported` is the list of the versions of the artifacts
+agent installed on the device can handle
```
kacf: `artifact_versions_supported` is not in your list above, and I think it is redundant, because the `version` header already has this number in it. It can be present in the Provides section, however.

Documentation/artifact-format.md
```diff
@@ -175,8 +273,89 @@ version is specified outside of `header.tar.gz`.
...
-### header-info
+#### device_provides
+
+There is also a set of parameters that are provided by the device itself,
+which are not a part of the artifact. Those are the values, that the device
+itself can read and send to the Mender server when needed. The full list of
+`device_provides` is as follows:
+
+* `device_type` is the current device type
+
```
kacf: I guess you need to add some entries here...

Documentation/artifact-format.md
```diff
@@ -175,8 +273,89 @@ version is specified outside of `header.tar.gz`.
...
+The `artifact_provides` is a set of parameters given artifact provides.
+For the detailed information see the description of the given parameter below.
+
+* `artifact_name` is the name of the artifact
+* `artifact_group` is the name of the group of artifacts given artifact
+belongs to
+* `rootfs_image_checksum` is the checksum of the image currently installed on the
+device
+* `update_types_supported` is the list of all the update types given Mender
+client can install
```
kacf: Did we want to restrict certain Provide types to the `type-info` file? Such as `rootfs_image_checksum` which is quite specific? I forget what we decided here.

Documentation/artifact-format.md
```diff
@@ -232,8 +411,41 @@ or multiple files listed.  For example:
...
+tar file. The first and the only required entry is the type of the update
+corresponding to the type in `header-info` file.
+It can also contain some additional parameters extending or modifying the global
+`artifact_provides` set of parameters specific for a given update type.
+
+```
+{
+  "type": "rootfs-image"
+  "artifact_provides": [
+          { "provides_type": ["provides_value", "provides_value"] },
```
kacf: Maybe some more entries here as well, depending on what we decide to put in each header file.

Documentation/artifact-format.md
```diff
@@ -332,6 +544,20 @@ For more information about the script and state API, see the official Mender
...
+
+Format: tar
+
+This file is complementing the information contained in the header.tar.gz.
+It can have the same structure as header.tar.gz, but for security reasons
+(this file is not signed) only certain files and parameters are allowed.
+
+At the moment ONLY header-info and type-info files are allowed which can
+contain only `artifact_provides`, `artifact_depends` `rootfs_image_checksum`
+parameters and type of the update.
```
kacf: I think only `artifact_depends` or? Because the end result is supposed to be identical, so `artifact_provides` can not be different.

Documentation/artifact-format.md
pasinskim: Disagree. You don't need to have a signature at all. Yet still you will use some "extra" parameters in those augment files.

Documentation/artifact-format.md
pasinskim: Fixed!

Documentation/artifact-format.md
pasinskim: But don't we need this for depends? For provides we will use the `version`, but we need to have some kind of information if the given artifact can be installed (if depends only on certain version).

Documentation/artifact-format.md
pasinskim: So this we will restrict in augment files only. I think here you can use whatever makes sense.

Documentation/artifact-format.md
pasinskim: You mean examples like `vexpress` or `beaglebone`?

Documentation/artifact-format.md
pasinskim: I've added `rootfs_image_checksum`.

Documentation/artifact-format.md
pasinskim: Right!

pasinskim:
@kacf can you recheck and comment? Thanks!

PR #109, opened 2018-01-31
MEN-1470 Fix ECDSA failures while signing and verifying artifact.

pasinskim:
In some cases the size of ECDSA keys is different than expected
32 bits. In those cases we've been storing both r and s keys in one
serialized slice (first half was s and the other r). In case of shorter
key we've been storing the key starting from the beginning of the slice
leaving trailing parts of the slice zeroed. This was causing issues
while reading the key and deserializing the whole 32 bits of the stored
slice.

Now we are using offset to leave the leading zeros of the buffer. This
is having no impact on storing and reading the serialized keys.

Changelog: Title

Signed-off-by: Marcin Pasinski <marcin.pasinski@northern.tech>
(cherry picked from commit 2119dcd2e3a2ad97eac2f6cf9354b966b15336bc)

### DIFF
--- a/artifact/signer.go
+++ b/artifact/signer.go
@@ -110,9 +110,27 @@ func (e *ECDSA256) Sign(message []byte, key interface{}) ([]byte, error) {
 	// half is the r and the other one s;
 	// as both values are ecdsa256curveBits size we need
 	// 2*ecdsa256keySize size slice to store both
+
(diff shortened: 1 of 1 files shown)


kacf:
Old and outdated. We are not planning any more releases of 2.0 anyway.

PR #111, opened 2018-02-02
 MEN-1742: ED25519 support in mender-artifact sign & verification

alfrunes:
Changelog: title


kacf:
We decided not to do this for now, so closing, but the branch is still here if we want to pick it up later.

PR #113, opened 2018-03-16
CLI refactoring  

pasinskim:


cli/mender-artifact/copy.go
```diff
@@ -107,12 +107,9 @@ func Copy(c *cli.Context) (err error) {
 }
 
 func Append(c *cli.Context) error {
-	fmt.Println("Hello")
```
kacf: Lol

cli/mender-artifact/copy.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+func parseCLIOptions(c *cli.Context) int {
+
+	finfo, err := os.Stdin.Stat()
+	if err != nil {
+		return criterror
+	}
+
+	isimg := regexp.MustCompile(`\.(mender|sdimg)`)
```
kacf: Same here, add `uefiimg`.

cli/mender-artifact/copy.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// no data on stdin
+	if finfo.Mode()&os.ModeNamedPipe == 0 {
```
kacf: I don't understand this, why are we checking for pipe status? Maybe a question for @oleorhagen.

Makefile
```diff
@@ -86,7 +86,7 @@ coverage:
 		rm -f coverage-tmp.txt;  \
 		$(GO) test -coverprofile=coverage-tmp.txt $$p ; \
 		if [ -f coverage-tmp.txt ]; then \
-			cat coverage-tmp.txt |grep -v 'mode:' >> coverage.txt; \
+			cat coverage-tmp.txt |grep -v 'mode:' | cat >> coverage.txt; \
```
kacf: Why? This is a noop.

cli/mender-artifact/log.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Log is a global reference to our logger.
+var Log *logrus.Logger
```
kacf: Do we want to use `mendersoftware/log` instead? It never turned out to be the great logger we thought of, so I'm alright if we don't. But the client uses it and it gives the file that the message came from. It's interface compatible with logrus.

cli/mender-artifact/main.go
```diff
@@ -0,0 +1,236 @@
...
+		Description: "Cat can output a file from a mender artifact or mender image to stdout.",
+		Action:      Cat,
+	}
+
+	append := cli.Command{
+		Name:        "append",
+		Usage:       "append [artifact|sdimg]:<filepath> <src>",
+		Description: "append or prepend to  file on a mender artifact or sdimg, where the placement of the '-' decides the append or prepend functionality",
+		Action:      Append,
+	}
```
kacf: Should mention `uefiimg` in all the usage strings as well.

cli/mender-artifact/main.go
```diff
@@ -0,0 +1,236 @@
...
+		Name:        "cat",
+		Usage:       "cat [artifact|sdimg]:<filepath>",
+		Description: "Cat can output a file from a mender artifact or mender image to stdout.",
+		Action:      Cat,
+	}
+
+	append := cli.Command{
+		Name:        "append",
+		Usage:       "append [artifact|sdimg]:<filepath> <src>",
+		Description: "append or prepend to  file on a mender artifact or sdimg, where the placement of the '-' decides the append or prepend functionality",
```
kacf: It is not clear from the description where '-' fits into this picture. The description mentions it, but the usage string doesn't contain it, so where does it go? @oleorhagen

cli/mender-artifact/copy.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	for i, arg := range c.Args() {
+		if arg == "-" {
+			appendint = i
+			break
+		}
+	}
+
+	isImg := regexp.MustCompile(`\.(mender|sdimg)`)
```
kacf: Could you add `uefiimg` to this while you're at it? This will become relevant very soon, and the code for producing such an image is already in meta-mender/master.

kacf:
Btw, you will also need to rebase, because I have made followup changes in mendertesting and its vendored dependency.

cli/mender-artifact/copy.go
oleorhagen: It is to see wether there is any data available on stdin using finfo see #220

cli/mender-artifact/copy.go
oleorhagen: No need. I’ve decided to remove the append functionality.

cli/mender-artifact/main.go
oleorhagen: Same as above.

cli/mender-artifact/main.go
oleorhagen: Probably. I’ll fix it up tomorrow

oleorhagen:
@pasinskim There is a PR open for your artifacts report that removes the append functionality and also significantly increases test coverage for my part of the PR.

cli/mender-artifact/copy.go
oleorhagen: Yeh. Embarrassing, but outdated thank god 🙈🙈

oleorhagen:
The uefiimg part is not in there though. But then tests should probably be added for it too?

kacf:
> The uefiimg part is not in there though. But then tests should probably be added for it too?

I don't think we need explicit tests for uefiimg regarding the artifact modification, since the format is practically identical to sdimg. It is the content which is different.

Makefile
```diff
@@ -34,7 +37,7 @@ BUILDTAGS = -tags '$(TAGS)'
 endif
 
 build:
-	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
+	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
```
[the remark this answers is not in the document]
pasinskim: Fixing!

Makefile
kacf: Oh! It's because grep fails? Clever!

cli/mender-artifact/main.go
kacf: We will, see my earlier comment.

kacf:
Another thing I'd like to ask here, is that there is a pretty drastic reduction in test coverage resulting from this pull request (almost 10%). Any idea why?

pasinskim:
@kacf I have cherry-picked some outstanding commits and run through building and testing the PR. I think all the comments should be addressed now. The build was green and test coverage is just slightly down. Considering the size of the PR, and the number of changes I would merge it and continue refactoring on top of master instead of here.

pasinskim:
Merge together with https://github.com/mendersoftware/meta-mender/pull/503

pasinskim:
Seems that the build was green https://mender-jenkins.mender.io/job/mender-builder/166/

PR #118, opened 2018-06-15
[WIP]:MEN-1672 - Add a support for calculating final image checksum

oleorhagen:

### DIFF
--- a/handlers/delta_image.go
+++ b/handlers/delta_image.go
@@ -1,0 +1,62 @@
+// Copyright 2018 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+package handlers
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"io"
+	"os"
+
+	"github.com/pkg/errors"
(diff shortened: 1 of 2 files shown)


handlers/delta_image.go
```diff
@@ -0,0 +1,62 @@
...
+			return errors.Wrap(err, "delta-installation failed")
+		}
+	}
+	// Get the checksum of the newly installed partition
+	partf, err := os.OpenFile(d.IPart, os.O_RDONLY, os.ModeDevice)
+	h := sha256.New()
+	if _, err = io.Copy(h, partf); err != nil {
+		return errors.Wrapf(err, "failed to calculate the checksum of the inactive partition")
+	}
+	if !bytes.Equal(h.Sum(nil), d.update.Checksum) {
```
oleorhagen: This will only be correct given that the checksum for the installed image is passed in to the DataFile struct that is embedded in the Rootfs struct

pasinskim: It should be like that, though I think you will need to change the way the checksum is validated while installing the image. See https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format.md#manifest-augment.

kacf: This task is listed below several other tasks, most notably [MEN-1668](https://tracker.mender.io/browse/MEN-1668). The way I envisioned it was that we would implement the new format with Provides and Depends first, and then use the new format while implementing this task. The implementation here will probably be useful, but it's a bit pointless in itself without the format changes that need the checksum verification.

oleorhagen: I disagree. This is a pretty freestanding solution, and the way I have read the code, something similar will have to be done anyways. There is no point in merging it yet of course, but still. If the entire installer solution have to be rewritten then you are right, but I don't think that will be necessary by the looks of it.

kacf: I agree this should probably live on a feature branch for the time being. And I suspect that we will need more integration with the new format, especially the augment files. But otherwise you are right, the actual verification should be pretty much like this. I think we are more or less on the same page then.

PR #120, opened 2018-06-19
Install function added 

oleorhagen:
Changelog: Title

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -108,13 +108,51 @@ func doCopy(c *cli.Context, repack *bool, r *io.ReadCloser,
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
 	case parseError:
-		return cli.NewExitError(fmt.Sprintln("no artifact or image provided"), 1)
+		return cli.NewExitError(fmt.Sprintln("no artifact or sdimage provided"), 1)
 	case argerror:
 		return cli.NewExitError(fmt.Sprintf("got %d arguments, wants two", c.NArg()), 1)
 	default:
 		return cli.NewExitError("critical error", 1)
 	}
 	return nil
+}
+
+// Install installs a file from the host filesystem onto either
+// a mender artifact, or an sdimg.
+func Install(c *cli.Context) (err error) {
+	var r io.ReadCloser
+	var w PartitionReadWriteClosePacker
+	switch parseCLIOptions(c) {
+	case copyin:
+		var perm os.FileMode
+		if c.Int("mode") == 0 {
+			return cli.NewExitError("File permissions needs to be set, if you are simply copying, the cp command should fit your needs", 1)
+		}
+		perm = os.FileMode(c.Int("mode"))
+		r, err = os.OpenFile(c.Args().First(), os.O_RDWR, perm)
+		defer r.Close()
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		w, err = NewPartitionFile(c.Args().Get(1), c.String("key"))
+		defer w.Close()
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		if _, err = io.Copy(w, r); err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		if err = w.Repack(); err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		return nil
+	case parseError:
+		return cli.NewExitError("No artifact or sdimg provided", 1)
+	case argerror:
+		return cli.NewExitError(fmt.Sprintf("got %d arguments, wants two", c.NArg()), 1)
+	default:
+		return cli.NewExitError("Unrecognized parse error", 1)
+	}
 }
 
 // enumerate cli-options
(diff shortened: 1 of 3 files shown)


cli/mender-artifact/main.go
```diff
@@ -215,6 +215,20 @@ func run() error {
 		Action:      Cat,
 	}
 
+	install := cli.Command{
+		Name:        "install",
+		Usage:       "install -m<permissions> <src> <dst>",
```
oleorhagen: Since this can only copy into the artifact/image, maybe a different usage description would be better. Any suggestions?

cli/mender-artifact/copy.go
```diff
@@ -117,6 +118,48 @@ func doCopy(c *cli.Context, repack *bool, r *io.ReadCloser,
...
+		if c.String("mode") == "" {
+			return cli.NewExitError("File permissions needs to be set, if you are simply copying, the cp command should fit your needs", 1)
+		}
+		var i int
+		i, err = strconv.Atoi(c.String("mode"))
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		perm = os.FileMode(i)
+		r, err = os.OpenFile(c.Args().First(), os.O_RDWR, perm)
```
kacf: I'm confused. Why are you using `perm` on a file you intend to read? It has no effect. And mode should be `O_RDONLY` for a file you will only read.

cli/mender-artifact/copy.go
```diff
@@ -117,6 +118,48 @@ func doCopy(c *cli.Context, repack *bool, r *io.ReadCloser,
...
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		w, err = NewPartitionFile(c.Args().Get(1), c.String("key"))
+		if err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		if _, err = io.Copy(w, r); err != nil {
+			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+		}
+		if err = w.Repack(); err != nil {
```
kacf: I would think that between `Repack` and the previous `if`, you should have:
1. An explicit `w.Close()`. Buffering can cause `w` to not be fully up to date from `debugfs`'s perspective unless you either flush or close the filehandle.
2. `Fchmod` on the `w` handle, which is where you set the permissions. Am I correct that `debugfs` grabs the permissions from the file you try to put in the image?

cli/mender-artifact/main.go
kacf: I think the description covers it.

cli/mender-artifact/copy_test.go
```diff
@@ -130,15 +131,36 @@ func TestCopy(t *testing.T) {
...
+			argv: []string{"mender-artifact", "install", "-m", "0600", "testkey", ":/etc/mender/testkey.key"},
+			verifyTestFunc: func(imgpath string) {
+				cmd := exec.Command(filepath.Join(dir, "mender-artifact"), "cat", imgpath+":/etc/mender/testkey.key")
+				var out bytes.Buffer
+				cmd.Stdout = &out
+				err := cmd.Run()
+				assert.Nil(t, err, "got unexpected error: %v", err)
+				assert.Equal(t, "foobar", out.String())
+				// Cleanup the testkey
+				assert.Nil(t, os.Remove("testkey"))
```
kacf: Shouldn't you test the permissions here as well, in addition to the contents?

cli/mender-artifact/copy.go
oleorhagen: The permission bits are copied over from debugfs

cli/mender-artifact/copy.go
oleorhagen: 1) debugfs has long since finished it's magic, when close is called on w. The close simply closes the tmp-files that are buffering the debugfs-data.
2) That is correct!

cli/mender-artifact/copy_test.go
oleorhagen: yes, sir!

cli/mender-artifact/copy_test.go
```diff
@@ -143,15 +143,54 @@ func TestCopy(t *testing.T) {
...
+				pf, err := NewPartitionFile(imgpath+":/etc/mender/testkey.key", "")
+				require.Nil(t, err)
+				// Type switch on the artifact, or sdimg underlying
+				switch pf.(type) {
+				case *partitionFile:
+					imgpath := pf.(*partitionFile).path
+					cmd := exec.Command("debugfs", "-R", "stat /etc/mender/testkey.key", imgpath)
+					out, err := cmd.CombinedOutput()
+					require.Nil(t, err)
+					require.True(t, strings.Contains(string(out), "Mode:  0600"))
```
oleorhagen: @kacf tests for mode set on the file

cli/mender-artifact/copy.go
```diff
@@ -117,6 +117,44 @@ func doCopy(c *cli.Context, repack *bool, r *io.ReadCloser,
...
+	var w PartitionReadWriteClosePacker
+	switch parseCLIOptions(c) {
+	case copyin:
+		var perm os.FileMode
+		if c.Int("mode") == 0 {
+			return cli.NewExitError("File permissions needs to be set, if you are simply copying, the cp command should fit your needs", 1)
+		}
+		perm = os.FileMode(c.Int("mode"))
+		r, err = os.OpenFile(c.Args().First(), os.O_RDWR, perm)
+		defer r.Close()
```
oleorhagen: Close added

cli/mender-artifact/copy.go
kacf: Hmm, ok, I think it's obvious that I'm not following the logic here. I'll trust you then! :-)

oleorhagen: I don't like that. Let's have a quick 2 min on it tomorrow.

PR #123, opened 2018-06-25
Checksums for testify/require update

oleorhagen:
Changelog: Title


oleorhagen:
Changed the testify/require and testify/assert dependencies to match (there was some inconsistencies with the versions used).

oleorhagen:
testify now matches the version used in the mender repo

PR #126, opened 2018-06-29
Refactored the partition image-file reader and writer implementation

oleorhagen:


cli/mender-artifact/partition.go
```diff
@@ -26,25 +27,106 @@ import (
...
+
+func stripSDimgBootExtensions(fpath string) string {
+	// Map /boot/{efi,grub} to /boot
+	if strings.HasPrefix(fpath, "/boot/efi/") {
+		return strings.TrimPrefix(fpath, "/boot/efi/")
+	} else if strings.HasPrefix(fpath, "/boot/grub") {
+		return strings.TrimPrefix(fpath, "/boot/grub/")
+	}
+	// /boot prefix
+	return strings.TrimPrefix(fpath, "/boot/")
```
kacf: Shouldn't this be `/uboot`? `/boot` is not the boot partition, it is the `boot` directory on the root partition. The three possibilities are:

* `/uboot`
* `/boot/efi`
* `/boot/grub`

There are a lot of places in the code this might need changing.

cli/mender-artifact/copy_test.go
```diff
@@ -143,18 +147,63 @@ func TestCopy(t *testing.T) {
 				assert.Equal(t, "artifact_name=foobar", out.String())
 			},
 		},
+		{
+			name: "write and read from the boot partition",
+			initfunc: func() {
+				require.Nil(t, ioutil.WriteFile("foo.txt", []byte("foobar"), 0644))
+			},
+			argv: []string{"mender-artifact", "cp", "foo.txt", ":/boot/test.txt"},
```
kacf: Should be `/uboot`.

PR #128, opened 2018-07-11
Tb

oleorhagen:


cli/mender-artifact/partition.go
```diff
@@ -230,131 +196,149 @@ func imgFilesystemType(imgpath string) (int, error) {
...
+	tmpf, err := ioutil.TempFile("", "mendertmp")
+	if err != nil {
+		return nil, err
+	}
+	// Strip the possible xtra boot paths
+	reg := regexp.MustCompile("/(uboot|boot/(efi|grub))")
+	// Only return the boot-partition.
+	if reg.MatchString(fpath) {
+		return nil, errors.New("newArtifactExtFile: A mender artifact does not contain a boot partition, only a rootfs")
+	}
```
kacf: What about data partition? Shouldn't that have an error as well?

PR #129, opened 2018-07-24
Updated the JSON format of header-info version 3.

oleorhagen:
Changelog: Title

### DIFF
--- a/Documentation/artifact-format.md
+++ b/Documentation/artifact-format.md
@@ -281,27 +281,29 @@ Format: JSON
 
 ```
 {
-  "updates": [
-    {
-      "type": "rootfs-image"
+    "updates": [
+        {
+            "type": "rootfs-image"
+        },
+        {
+            "type": "delta-image"
+        }
+    ],
+    "artifact_provides": {
+        "artifact_name": "release-2",
+        "artifact_group": "fix",
+        "update_types_supported": [
+            "rootfs-image",
+            "delta-update"
+        ]
     },
-    {
-      "type": "rootfs-image"
-    },
-    {
-      ...
+    "artifact_depends": {
+        "artifact_name": "release-1",
+        "device_type": [
+            "vexpress-qemu",
+             "beaglebone"
+        ]
     }
-  ],
-  "artifact_provides": [
-          { "artifact_name": "name" },
-          { "artifact_group": "group-1" },
-          { "update_types_supported": ["rootfs-image"]},
-      ],
-  "artifact_depends": [
-          { "device_type": ["vexpress-qemu", "beaglebone"] },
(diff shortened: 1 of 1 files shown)


Documentation/artifact-format.md
```diff
@@ -281,29 +281,30 @@ Format: JSON
...
-      "type": "rootfs-image"
+    "updates": [
+        {
+            "type": "rootfs-image"
+        },
+        {
+            "type": "delta-image"
+        }
+    ],
+    "ArtifactProvides": {
```
oleorhagen: Shouldn't artifact-provides be an object? Or should this be a list of artifact-provides?

Documentation/artifact-format.md
```diff
@@ -281,29 +281,30 @@ Format: JSON
...
+        "artifact_name": "release-2",
+        "artifact_group": "fix",
+        "update_types_supported": [
+            "rootfs-image",
+            "delta-update"
+        ]
     },
-    {
-      ...
+    "ArtifactDepends": {
```
oleorhagen: Same as above

Documentation/artifact-format.md
```diff
@@ -281,29 +281,30 @@ Format: JSON
...
+        "artifact_name": "release-1",
+        "device_type": [
+            "vexpress-qemu",
+             "beaglebone"
+        ]
     }
-  ],
-  "artifact_provides": [
-          { "artifact_name": "name" },
```
oleorhagen: It's not necessary for every parameter to be an object is it?

kacf: Correct me if I'm wrong, but I think that inside a list they have to be.

Documentation/artifact-format.md
```diff
@@ -281,29 +281,30 @@ Format: JSON
...
-          { "update_types_supported": ["rootfs-image"]},
-      ],
-  "artifact_depends": [
-          { "device_type": ["vexpress-qemu", "beaglebone"] },
-          { "artifact_name": ["rootfs-1"] },
-      ],
-  ],
-}
-```
+}```
```
kacf: It's usual markdown style to have the triple backticks on their own line. Not sure if the rendering is affected, but just convention.

Documentation/artifact-format.md
kacf: Ah, but you've turned the whole thing from a list to an object... Hmm that might work too, and maybe that's better, since each entry is a key/value pair. Never mind then!

Documentation/artifact-format.md
kacf: Oh, and the rendering *is* affected, I see now!

oleorhagen:
@kacf backticks now on a seperate line

PR #130, opened 2018-07-30
[WIP] V3.1

oleorhagen:


oleorhagen:
Not finished, but what is done, and should probably be looked at: 
* HeaderInfoV3
* TypeInfoV3
* WriteArtifact
* ReadArtifact
* Api refactorization into parameter structs

Ignore the cli modifications for now. Also, although there are many comits, I would recommend looking at this in one blob.

What is missing:
- header and manifest augment.
- cli (Should this be added?)

areader/reader.go
```diff
@@ -221,14 +281,15 @@ func signatureReadAndVerify(tReader *tar.Reader, message []byte,
 	if verify == nil && signed {
 		return errors.New("reader: verify signature callback not registered")
 	} else if verify != nil {
+		fmt.Println("signatureReadAndVerify: ")
```
pasinskim: Probably you don't want to leave it here, do you?

oleorhagen:
Mind you, most of the lines are tests though.

oleorhagen:
@pasinskim I propose to make this into one PR for the updated:
* HeaderInfoV3
*TypeInfoV3
*WriteArtifactV3

And then make a task to rewrite the parsing of all the artifact versions.

oleorhagen:
If anyone feels like finishing this while I'm gone, have a look at the message in the last commit.

PR #133, opened 2018-08-03
FIX: remove extraneous files from using mender-artifact cp tool. 

oleorhagen:
Changelog: None

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -46,30 +46,34 @@ func Cat(c *cli.Context) (err error) {
 func Copy(c *cli.Context) (err error) {
 	var r io.ReadCloser
 	var w io.WriteCloser
-
 	switch parseCLIOptions(c) {
 	case copyin:
-		r, err = os.OpenFile(c.Args().First(), os.O_CREATE|os.O_RDWR, 0655)
+		r, err = os.OpenFile(c.Args().First(), os.O_CREATE|os.O_RDONLY, 0655)
+		defer r.Close()
 		if err != nil {
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
 		w, err = NewPartitionFile(c.Args().Get(1), c.String("key"))
 		if err != nil {
+			w.Close()
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
 	case copyinstdin:
 		r = os.Stdin
 		w, err = NewPartitionFile(c.Args().First(), c.String("key"))
 		if err != nil {
+			w.Close()
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
 	case copyout:
 		r, err = NewPartitionFile(c.Args().First(), c.String("key"))
+		defer r.Close()
 		if err != nil {
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
-		w, err = os.OpenFile(c.Args().Get(1), os.O_CREATE|os.O_RDWR, 0655)
+		w, err = os.OpenFile(c.Args().Get(1), os.O_CREATE|os.O_WRONLY, 0655)
 		if err != nil {
+			w.Close()
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
 	case parseError:
@@ -81,14 +85,11 @@ func Copy(c *cli.Context) (err error) {
 	}
 
 	_, err = io.Copy(w, r)
-	if err != nil {
-		return cli.NewExitError(err.Error(), 1)
+	// Ignore the close error in case of an io.Copy error.
+	if cerr := w.Close(); cerr != nil && err != nil {
+		err = cerr
 	}
-
-	if err = w.Close(); err != nil {
-		return err
-	}
-	return r.Close()
+	return err
 }
 
 // Install installs a file from the host filesystem onto either
--- a/cli/mender-artifact/copy_test.go
+++ b/cli/mender-artifact/copy_test.go
@@ -196,6 +196,7 @@ func TestCopy(t *testing.T) {
 				assert.Nil(t, os.Remove("testkey"))
 				// Check that the permission bits have been set correctly!
 				pf, err := NewPartitionFile(imgpath+":/etc/mender/testkey.key", "")
+				defer pf.Close()
 				require.Nil(t, err)
 				// Type switch on the artifact, or sdimg underlying
 				switch pf.(type) {
--- a/cli/mender-artifact/partition.go
+++ b/cli/mender-artifact/partition.go
@@ -42,6 +42,12 @@ type partition struct {
 type sdimgFile []io.ReadWriteCloser
 
 func newSDImgFile(fpath string, modcands []partition) (sdimgFile, error) {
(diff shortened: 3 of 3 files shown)


oleorhagen:
This removes all leftover files in /tmp from using mender-artifact cp, cat or install.

cli/mender-artifact/artifacts.go
```diff
@@ -104,7 +104,7 @@ func unpackArtifact(name string) (string, error) {
 	aReader := areader.NewReader(f)
 	rootfs := handlers.NewRootfsInstaller()
 
-	tmp, err := ioutil.TempFile(filepath.Dir(name), "mender-artifact")
```
pasinskim: Why have you changed this? Now the temporary file will be created in `/tmp` instead of the same location where the artifact is. I think the reason for storing it in the same location was that you might not have access to `/tmp` (I am not sure if this was the case for Travis), but you most probably can create some temporary files in your work directory.

cli/mender-artifact/copy.go
```diff
@@ -46,6 +46,14 @@ func Cat(c *cli.Context) (err error) {
...
 	var r io.ReadCloser
 	var w io.WriteCloser
+	defer func() {
+		if r != nil {
+			r.Close()
+		}
+		if w != nil {
+			w.Close()
+		}
+	}()
```
pasinskim: Should we try to return the error here?

cli/mender-artifact/copy.go
```diff
@@ -85,10 +93,7 @@ func Copy(c *cli.Context) (err error) {
 		return cli.NewExitError(err.Error(), 1)
 	}
 
-	if err = w.Close(); err != nil {
-		return err
-	}
-	return r.Close()
+	return nil
```
pasinskim: If we will assign the 'err' named return variable correctly we should skip this one.

cli/mender-artifact/copy_test.go
```diff
@@ -196,6 +196,7 @@ func TestCopy(t *testing.T) {
 				assert.Nil(t, os.Remove("testkey"))
 				// Check that the permission bits have been set correctly!
 				pf, err := NewPartitionFile(imgpath+":/etc/mender/testkey.key", "")
+				defer pf.Close()
```
pasinskim: Should we also remove it afterwords?

cli/mender-artifact/partition.go
```diff
@@ -42,13 +42,20 @@ type partition struct {
 type sdimgFile []io.ReadWriteCloser
 
 func newSDImgFile(fpath string, modcands []partition) (sdimgFile, error) {
+	// NOTE: Closure that captures the partitons-to-be-deleted variable,
+	// and cleans ups all the unused partitions on exit.
+	var delPartition []int = nil // By default clean up everything.
```
pasinskim: Is it removing all by default? It seems that it removes nothing if `delPartition` is empty.

cli/mender-artifact/partition.go
```diff
@@ -105,9 +110,23 @@ func newSDImgFile(fpath string, modcands []partition) (sdimgFile, error) {
...
 		rootfsb,
 	}
+	delPartition = []int{0, 3}
 	return ps, nil
 }
 
+// cleanupUnusedPartitions removes the unused tmpfiles from /tmp,
+// but does not report any errors.
+// usedPartition: nil -- cleans up all the tmp-files given.
+func cleanupUnusedPartitions(partitions []partition, unusedPartition ...int) {
```
pasinskim: I don't like how you are passing the `unusedPartition`. This indicates you are expecting multiple `int` parameters, which is OK, but in fact, you are passing a slice with the numbers of partitions to delete. What is more, why can't you store a list with the paths of the files to delete? We will eliminate looping and this info is accessible from where you are calling this function.

cli/mender-artifact/partition.go
```diff
@@ -202,8 +221,16 @@ type artifactExtFile struct {
...
+func newArtifactExtFile(key, fpath string, p partition) (af *artifactExtFile, err error) {
 	tmpf, err := ioutil.TempFile("", "mendertmp")
+	// Cleanup resources in case of error.
+	defer func() {
+		if err != nil {
+			tmpf.Close()
+			os.Remove(tmpf.Name())
+			os.Remove(p.path)
+		}
+	}()
```
pasinskim: Shouldn't this be rather handled by the caller of the function? 
```
if aFile, err := newArtifactExtFile(); err != nil {
    aFile.Close()
}
```

cli/mender-artifact/partition.go
```diff
@@ -225,7 +252,16 @@ func newArtifactExtFile(key, fpath string, p partition) (*artifactExtFile, error
...
+	if a == nil {
+		return nil
+	}
+	defer func() {
+		if a.tmpf != nil {
+			a.tmpf.Close()
+			os.Remove(a.tmpf.Name())
+		}
+	}()
+	defer os.Remove(a.path)
```
pasinskim: Can't we have one defer function?

cli/mender-artifact/partition.go
```diff
@@ -242,8 +278,16 @@ type extFile struct {
...
+func newExtFile(key, imagefilepath string, p partition) (e *extFile, err error) {
 	tmpf, err := ioutil.TempFile("", "mendertmp")
+	// Cleanup resources in case of error.
+	defer func() {
+		if err != nil {
+			tmpf.Close()
+			os.Remove(tmpf.Name())
+			os.Remove(p.path)
+		}
+	}()
```
pasinskim: As above. I would prefer the cleanup to be done by the caller.

cli/mender-artifact/partition.go
```diff
@@ -274,24 +318,28 @@ func (ef *extFile) Write(b []byte) (int, error) {
...
+	if ef == nil {
+		return nil
+	}
 	if ef.repack {
 		part := []partition{ef.partition}
 		err = repackSdimg(part, ef.name)
 	}
-	ef.tmpf.Close()    // Ignore
+	ef.tmpf.Close() // Ignore
+	os.Remove(ef.tmpf.Name())
```
pasinskim: Should you somehow store `err` here as well?

cli/mender-artifact/partition.go
```diff
@@ -304,6 +352,18 @@ type fatFile struct {
 	tmpf          *os.File
 }
 
+func newFatFile(imageFilePath string, partition partition) (*fatFile, error) {
+	tmpf, err := ioutil.TempFile("", "mendertmp")
```
pasinskim: Suggestion: Maybe we could stick to some convention - mendertmp-FILENAME or something to track better the resource leaks if happening?

oleorhagen: Great idea! simple and easy. Like it.

cli/mender-artifact/copy_test.go
oleorhagen: No, the `PartitionFile` does only exist temporarily, so there is no point in removing it. A close will do.

oleorhagen:
@pasinskim I have now fixed up all the review comments, and the tests are passing locally. However - Travis keeps failing, with something that looks to me like a QA Issue, as the go vet tool panics as it is walking the syntax tree.

PR #138, opened 2018-09-10
Mender artifact v3 

oleorhagen:


oleorhagen:
@pasinskim This is the current progress on the Mender-artifact V3 format, I know that it is huge, but I have tried breaking it down in bite-sized chunks for you, by factoring the functionality into different commits.
Travis will fail as the `handleHeaderReads` function in areader/reader.go has a cyclomatic complexity of 18. I have not broken it down further, as I do not think it will help readability, but I will let you be the judge of that.  The `TODO`'s are left in there as questions for you, as I am not sure what is the best, or required way this should function.

Hope this is bearable.

cli/mender-artifact/main.go
```diff
@@ -94,6 +94,22 @@ func run() error {
...
+		},
+		// TODO - this flag, or device-type?
+		cli.StringSliceFlag{
+			Name:  "artifact-device-depends",
+			Usage: "Sets the name(s) of the device(s) which this update depends upon",
+		},
+		cli.StringFlag{
+			Name:  "group",
+			Usage: "Sets the artifact group",
+		},
```
kacf: Two things:

* I would rename it to `provides-group`. It's still a nice sounding flag, and more similar to the spec, so less chance of confusion, if only among us.
* There should be a `depends-groups` as well, which should be a list aka `-t` which you can specify multiple times. This is not described super well in the spec (it's missing from the example), but you can see it in the list of members for artifact_depends.

cli/mender-artifact/write.go
```diff
@@ -15,36 +15,69 @@
...
-		)
-	}
-	if len(strings.Fields(c.String("artifact-name"))) > 1 {
-		// check for whitespace in artifact-name
-		return cli.NewExitError(
-			"whitespace is not allowed in the artifact-name",
-			errArtifactInvalidParameters,
-		)
+	if c.Int("version") == 3 {
+		// TODO - spec says that device-type can be empty.
```
kacf: It does? Where?

It's probably wrong. Device-type must be present.

areader/reader.go
```diff
@@ -245,10 +308,143 @@ func verifyVersion(ver []byte, manifest *artifact.ChecksumStore) error {
...
 }
 
+// The artifact parser needs to take one of the paths listed below when reading
+// the artifact version 3.
+var artifactV3ParseGrammar = [][]string{
+	// Version is already read in ReadArtifact().
+	{"manifest", "header.tar.gz"},                                                              // Unsigned.
+	{"manifest", "manifest.sig", "header.tar.gz"},                                              // Signed.
+	{"manifest", "manifest-augment", "header.tar.gz", "header-augment.tar.gz"},                 // TODO - is this the unsigned path, or the one above the preferred one?.
+	{"manifest", "manifest.sig", "manifest-augment", "header.tar.gz", "header-augment.tar.gz"}, // Augmented.
```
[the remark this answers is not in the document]
oleorhagen: Still, which path should be taken by an unsigned artifact?

areader/reader.go
```diff
@@ -245,10 +308,143 @@ func verifyVersion(ver []byte, manifest *artifact.ChecksumStore) error {
...
+			// We have a submatch. Check if the entire length matches.
+			if i == len(parseOrder)-1 {
+				if len(parseOrder) == len(validPath) {
+					return parseOrder[i], true // Full match.
+				}
+				return parseOrder[i], false
+			}
+		}
+	}
+	return fmt.Sprintf("Invalid structure: %s, wrong element: %s", parseOrder, parseOrder[len(parseOrder)-1]), false
```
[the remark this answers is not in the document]
oleorhagen: Agreed. Gonna fix this filth up!

artifact/metadata_test.go
```diff
@@ -34,29 +37,246 @@ func TestValidateInfo(t *testing.T) {
...
+		{HeaderInfo{Updates: []UpdateType{{Type: ""}}},
+			"Artifact validation failed with missing arguments: No compatible devices listed, No artifact name, Empty update"},
+		{HeaderInfo{Updates: []UpdateType{{Type: "update"}, {}}},
+			"Artifact validation failed with missing arguments: No compatible devices listed, No artifact name, Empty update"},
+		{HeaderInfo{Updates: []UpdateType{{}, {Type: "update"}}, CompatibleDevices: []string{""}, ArtifactName: "id"},
+			"Artifact validation failed with missing argument: Empty update"},
+		{HeaderInfo{Updates: []UpdateType{{Type: "update"}, {Type: ""}}, CompatibleDevices: []string{""}, ArtifactName: "id"},
+			"Artifact validation failed with missing argument: Empty update"},
+		{HeaderInfo{Updates: []UpdateType{{Type: "update"}}, CompatibleDevices: []string{""}, ArtifactName: "id"},
+			""},
```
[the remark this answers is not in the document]
oleorhagen: Leaving this fix for later, as we discussed.

areader/reader.go
```diff
@@ -114,23 +178,23 @@ func (ar *Reader) readHeader(tReader io.Reader, headerSum []byte) error {
...
 	// first part of header must always be header-info
-	hInfo := new(artifact.HeaderInfo)
+	hInfo := new(artifact.AugmentedHeaderInfoV3)
 	if err = readNext(tr, hInfo, "header-info"); err != nil {
 		return err
 	}
-	ar.hInfo = hInfo
+	ar.augmentedhInfo = hInfo
 
+	// TODO - necessary when the previous header is already processed?
```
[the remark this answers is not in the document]
oleorhagen: It does, and with the clarifications you made regarding the header vs header-augment, this is no longer an issue.

areader/reader_test.go
```diff
@@ -154,14 +186,17 @@ func TestReadArtifact(t *testing.T) {
...
 		err = aReader.ReadArtifact()
 		if test.readError != nil {
+			if err == nil {
+				log.Fatalf("%q: Test expected an error, but ReadArtifact did not return an error. expected: %v", name, test.readError)
+			}
 			assert.Equal(t, test.readError.Error(), err.Error())
 			continue
 		}
-		assert.NoError(t, err)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
oleorhagen: I use requires when there are errors from other code-modules, that are simply dependencies of the functionality, thus a failure in one of the dependent modules invalidates my tests, and I therefore abort immediately.

areader/reader.go
```diff
@@ -368,7 +359,13 @@ func (ar *Reader) readHeaderV3(tReader *tar.Reader,
...
 		}
 		parsePath = append(parsePath, hdr.Name)
-		nextParseToken, validPath := verifyParseOrder(parsePath)
+		nextParseToken, validPath, err := verifyParseOrder(parsePath)
+		if err != nil {
+			panic("The only error returned from verifyParseOrder should be errParseOrder")
+		}
+		if err == errParseOrder {
+			return nil, fmt.Errorf("Invalid structure: %s, wrong element: %s", parsePath, parsePath[len(parsePath)-1])
+		}
```
kacf: Hmm, having a `panic` makes me a bit nervous, even though it is an invariant check. This code will run both on the client and in the server. Could we simply return the same error instead?

In any case, you probably meant to reverse the ifs, since you will get the panic for all errors if they stay in this order.

cli/mender-artifact/write.go
kacf: Remove comment?

areader/reader.go
oleorhagen: :see_no_evil:  yeh, agreed.

oleorhagen: Btw, how do we then guarantee invariance? or should it simply be ignored?

areader/reader.go
```diff
@@ -340,8 +340,8 @@ func (ar *Reader) readHeaderV3(tReader *tar.Reader,
 	manifestChecksumStore := artifact.NewChecksumStore()
 	parsePath := []string{}
 
-	// Register the update handler for rootfsV3.
-	rootfs := handlers.NewRootfsInstaller(3)
+	// Register the update handler.
+	rootfs := handlers.NewRootfsInstaller()
```
oleorhagen: Sry for another bulk with a lot of files changing, but there really isn't much to it this time. 
1. I have changed the handler to not accept the version, and instead accept version as a parameter to the `handler.ReadHeader` function. This was required, as the client adds the install-handler before running `ReadArtifact`.

areader/reader.go
```diff
@@ -369,6 +369,11 @@ func (ar *Reader) readHeaderV3(tReader *tar.Reader,
 			return nil, errors.Wrap(err, "readHeaderV3")
 		}
 		if validPath {
+			// Artifact should be signed, but isn't, so do not process the update.
+			if ar.shouldBeSigned && !ar.IsSigned {
+				return nil,
+					errors.New("reader: expecting signed artifact, but no signature file found")
```
oleorhagen: 2. This was found running the client tests. What it means is that we should not proceed with the update, if the client provides a key, and the artifact is not signed. Had not thought of this, good thing we have tests! (Enjoy this while u can)

areader/reader.go
kacf: Well, we *do* guarantee invariance, it's just a different way of handling it. If we hit the check here then we return from the function, and outside the function the invariance is no longer relevant. We will thus fail to parse correctly, which is a grave bug in dire need of fixing, but at least your server won't crash and can (hopefully) still handle other duties.

Hard invariant asserts are in general a better fit for C, where such checks can expose that your memory is all messed up, which can be *really* bad. Truly messing up memory is much harder to achieve in Go, so a softer fallback can be used.

areader/reader.go
kacf: Haha, join the dark side! But yes, good catch indeed!

kacf:
> Sry for another bulk with a lot of files changing, but there really isn't much to it this time.

Not at all! It sits nicely in its own commit, so it's easy to review.

oleorhagen:
Well, the tests will keep failing as long as the cyclomatic complexity of the `handleheaderreads` > 15. However, I do not think that splitting it up will make it any easier to read! Do you still think we should refactor it? @kacf

kacf:
`handleHeaderReads` doesn't look so difficult to split. Can you just make a dedicated function for each of the cases, and use `handleHeaderReads` for the `switch` statement only?

Is it easier to read? Debatable. But not too hard to fix in this case. :-)

artifact/metadata.go
```diff
@@ -126,6 +311,42 @@ func (ti *TypeInfo) Write(p []byte) (n int, err error) {
...
+// TypeInfoV3 provides information about the type of update contained within the
+// headerstructure.
+type TypeInfoV3 struct {
+	// Rootfs/Delta (Required).
+	Type string `json:"type"`
+	// Checksum of the image that needs to be installed on the device in order to
+	// apply the current update.
+	ArtifactDepends []TypeInfoDepends `json:"artifact_depends,omitempty"`
+	// Checksum of the image currently installed on the device.
+	ArtifactProvides []TypeInfoProvides `json:"artifact_provides,omitempty"`
```
kacf: Am I correct that these two should be changed now after https://github.com/mendersoftware/mender-artifact/pull/139#discussion_r225507535 ?

oleorhagen: Well, if you mean that they should be changed to key-value pairs, I was a bit ahead of time (?), and thus this is the way it is already implemented.

oleorhagen: Well, if you mean that they should be changed to key-value pairs, I was a bit ahead of time (?), and thus this is the way it is already implemented.

oleorhagen: Well, if you mean that they should be changed to key-value pairs, I was a bit ahead of time (?), and thus this is the way it is already implemented.

oleorhagen: Well, if you mean that they should be changed to key-value pairs, I was a bit ahead of time (?), and thus this is the way it is already implemented.

oleorhagen:
@kacf I think this might actually be ready to go now. We'll see when the hooks are back up and running. There is a lot of commits I know, but mostly they are test-coverage fixes. The important ones are:
* https://github.com/mendersoftware/mender-artifact/pull/138/commits/12c4aca50e145d1f72684da083e7efc2181d4c04
*  https://github.com/mendersoftware/mender-artifact/pull/138/commits/040a72ec3f43ce28dcea0c550e2eaa9e14956bea
*  https://github.com/mendersoftware/mender-artifact/pull/138/commits/886a3b08f765c3e1dcdb875bbb1ab7a200d6c696

oleorhagen:
If you approve I will go ahead and squash some.

artifact/metadata.go
oleorhagen: yes

oleorhagen:
@kacf Woop, Woop! Now with only integration tests failing, I think we can agree that this is the base for mender-artifact-v3. Thus I will leave this as is, and further PR's will be directed to this branch. Time to close?

kacf:
Nice! Yeah, I think we can live with failing integration tests for the moment. No need to impair progress. So this can be closed, and the branch we keep!

I'll rebase my stuff on top of this as soon as I get the chance.

oleorhagen:
Superseeded by https://github.com/mendersoftware/mender-artifact/pull/180

PR #140, opened 2018-09-25
MEN-2155 Fail validate when key supplied for unsigned artifact

cosinekitty:
Changelog: A command of the form
"mender-artifact validate unsigned.mender -k public.key"
was incorrectly succeeding for an unsigned artifact when a public key
was supplied. Supplying a public key indicates that the caller requires
the artifact to contain a signature that matches that key.
Now this command fails (exits with a nonzero value) as expected.

Fixed logic in validate function to return an error when
a public key was given but we never received a callback to
validate a signature inside the artifact.

Added 2 more unit tests:

- Must fail validation of an unsigned artifact with key given.
  Confirmed this test failed before the fix mentioned above,
  and now passes.

- Must fail validation of a signed artifact with no key given.
  This test already, and still, passes.  It's just nice to have.

Changed TestValidate function to assert that the error is not nil
before asserting that it contains the expected text.
This makes a failure easier to understand when the error is
incorrectly nil, because it shows the line number of the assert
failure, instead of causing a panic before the assert is entered.
This happened to me when I ran the unit tests before fixing MEN-2155
and made it less obvious what was wrong.

### DIFF
--- a/cli/mender-artifact/validate.go
+++ b/cli/mender-artifact/validate.go
@@ -33,6 +33,7 @@ func validate(art io.Reader, key []byte) error {
 	// just continue checking consistency and return info if
 	// signature verification failed
 	var validationError error
+	verifiedDigitalSignature := false
 	verify := func(message, sig []byte) error {
 		verifyCallback := func(message, sig []byte) error {
 			return errors.New("artifact is signed but no verification key was provided")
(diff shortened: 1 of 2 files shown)


cosinekitty:
OK, clearly I broke something. I can see that the new `cat` command sends down the empty string "" as the key when there is no key specified, instead of `nil` like I expected. I will try again soon with another pull request. I'm still learning golang development; I just now learned how to run all the unit tests locally before submitting a pull request, so hopefully I won't make as much noise in the future.

kacf:
No worries, we've all been there! :-)

PR #141, opened 2018-09-25
MEN-2155 Fail validate when key supplied for unsigned artifact

cosinekitty:
Changelog: A command of the form
"mender-artifact validate unsigned.mender -k public.key"
was incorrectly succeeding for an unsigned artifact when a public key
was supplied. Supplying a public key indicates that the caller requires
the artifact to contain a signature that matches that key.
Now this command fails (exits with a nonzero value) as expected.

Fixed logic in validate function to return an error when
a public key was given but we never received a callback to
validate a signature inside the artifact.

Also fixed a problem where some callers (for example,
the new "cat" command) call validate with "" instead
of nil to indicate a missing key.

Added 2 more unit tests:

- Must fail validation of an unsigned artifact with key given.
  Confirmed this test failed before the fix mentioned above,
  and now passes.

- Must fail validation of a signed artifact with no key given.
  This test already, and still, passes.  It's just nice to have.

Changed TestValidate function to assert that the error is not nil
before asserting that it contains the expected text.
This makes a failure easier to understand when the error is
incorrectly nil, because it shows the line number of the assert
failure, instead of causing a panic before the assert is entered.
This happened to me when I ran the unit tests before fixing MEN-2155
and made it less obvious what was wrong.

### DIFF
--- a/cli/mender-artifact/validate.go
+++ b/cli/mender-artifact/validate.go
@@ -33,11 +33,15 @@ func validate(art io.Reader, key []byte) error {
 	// just continue checking consistency and return info if
 	// signature verification failed
 	var validationError error
+
+	// Some callers pass nil, others pass "", to indicate a missing public key.
+	keyIsSpecified := (key != nil) && (len(key) > 0)
+
 	verify := func(message, sig []byte) error {
 		verifyCallback := func(message, sig []byte) error {
 			return errors.New("artifact is signed but no verification key was provided")
 		}
-		if key != nil {
+		if keyIsSpecified {
 			s := artifact.NewVerifier(key)
 			verifyCallback = s.Verify
 		}
@@ -57,6 +61,10 @@ func validate(art io.Reader, key []byte) error {
 	}
 	if validationError != nil {
 		Log.Debugf("error validating signature: %s", validationError.Error())
+		return ErrInvalidSignature
+	}
+	if keyIsSpecified && !ar.IsSigned {
+		Log.Debug("key was specified but no digital signature was found")
 		return ErrInvalidSignature
 	}
 	return nil
--- a/cli/mender-artifact/validate_test.go
+++ b/cli/mender-artifact/validate_test.go
@@ -72,6 +72,8 @@ var validateTests = []struct {
 		ErrInvalidSignature},
 	{2, []byte(PrivateValidateRSAKey), []byte(PublicValidateRSAKeyInvalid),
 		ErrInvalidSignature},
+	{2, []byte(PrivateValidateRSAKey), nil, ErrInvalidSignature},
+	{2, nil, []byte(PublicValidateRSAKey), ErrInvalidSignature}, // MEN-2155
 }
 
 func TestValidate(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cli/mender-artifact/validate.go
```diff
@@ -33,18 +33,26 @@ func validate(art io.Reader, key []byte) error {
 	// just continue checking consistency and return info if
 	// signature verification failed
 	var validationError error
+	verifiedDigitalSignature := false
+
+	// Some callers pass nil, others pass "", to indicate a missing public key.
+	isKeySpecified := (key != nil) && (len(key) > 0)
```
oleorhagen: Just a question, doesn't keyIsSpecified` make the if statement sound a bit better?

cli/mender-artifact/validate.go
```diff
@@ -33,18 +33,26 @@ func validate(art io.Reader, key []byte) error {
...
+		if isKeySpecified {
 			s := artifact.NewVerifier(key)
 			verifyCallback = s.Verify
 		}
 
 		if verifyCallback != nil {
 			if err := verifyCallback(message, sig); err != nil {
 				validationError = err
+			} else {
+				verifiedDigitalSignature = true
```
oleorhagen: Maybe this can be recovered from the reader struct after a read is done? with
`reader.IsSigned` ?

cli/mender-artifact/validate.go
cosinekitty: I'm fine either way. `keyIsSpecified` does sound more like natural English. Beginning a boolean variable with "is" was just a holdover from a coding standard from a place I used to work. I'm happy to change it.

cosinekitty:
I like this idea. I didn't notice there was a Reader.IsSigned already.  I'm
puzzled why I don't see this comment on the pull request itself, like your
other comment. I wanted to reply to it there.

mirzak:
We normally rebase, squash and force push. Github is able to keep track of this, and will mark previous code/conversions as "outdated"

cosinekitty:
Thanks to @oleorhagen for the ideas. I think this is easier to understand now.
@mirzak I learned some new Git tricks!

cosinekitty:
I don't understand that test failure... is it an error uploading an artifact?

```
>       assert r.status_code == requests.status_codes.codes.created
E       AssertionError
```

mirzak:
Yeah, upload fails. Hardly believe it is related to your changes. I will look in to if I am able to rerun the tests.

cosinekitty:
FYI, I later noticed the same failure in another PR that only has documentation changes!
https://github.com/mendersoftware/mender-artifact/pull/139

mirzak:
Yeah, noticed that one as well. This probably will not be fixed until Monday, so sit tight :)

oleorhagen:
Hi, sry for the late response, I'm only here a couple of days a week. Yeh, the failure looks spurious to me as well, rerunning them now.

oleorhagen:
Alright, all good :)  
Thank you so much for your contribution, and a clean and really good looking PR!
@cosinekitty

PR #142, opened 2018-09-26
FIX: MEN-2166: Renaming a file across devices now works.

oleorhagen:
Changelog: None

### DIFF
--- a/cli/mender-artifact/sign.go
+++ b/cli/mender-artifact/sign.go
@@ -17,6 +17,7 @@ package main
 import (
 	"io/ioutil"
 	"os"
+	"path/filepath"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


oleorhagen:
Should also be backported to 2.0 and 2.1 then?

kacf:
Thanks!

This is the kind of change that should have a changelog entry btw. I noticed it after merging, but I added [an entry](https://github.com/mendersoftware/mender-artifact/commit/2995974a0541d6382fd83335b7df0e3018ffabc3) afterwards.

Will only [backport this to 2.3](https://github.com/mendersoftware/mender-artifact/pull/144), since 2.0 and 2.1 won't receive any more updates, and 2.2 doesn't have the bug.

PR #147, opened 2018-10-01
MEN-2169 Fix mender-artifact "cat", "cp", "modify" on signed artifacts.

cosinekitty:
Changelog: Fixed a bug that caused a command like
"mender-artifact cat signed.mender:/etc/mender/artifact_info"
to fail with the error:
"failed to open the partition reader: err: error validating signature"
There was a similar problem with the "cp" command, and also
the "modify" command when no "-k" was present to replace the
existing signature.

This change is based on the general idea that if the user
does not specify a public key, he is saying he does not want
to authenticate a digital signature, whether or not one exists
in the artifact.

Simplified the validate function so it no longer returns an
error if the artifact contains a signature but no public
key was specified by the caller.

Added a unit test that used to fail before the bug was fixed,
and now it passes.

Updated existing unit tests that were affected by this change.

### DIFF
--- a/cli/mender-artifact/modify_existing_test.go
+++ b/cli/mender-artifact/modify_existing_test.go
@@ -252,6 +252,7 @@ func TestModifySigned(t *testing.T) {
 	err = ioutil.WriteFile(filepath.Join(tmp, "ecdsa.key"), []byte(PrivateECDSAKey), 0711)
 	assert.NoError(t, err)
 
+	// Create and sign artifact using RSA private key.
 	os.Args = []string{"mender-artifact", "write", "rootfs-image", "-t", "my-device",
 		"-n", "release-1", "-u", filepath.Join(tmp, "mender_test.img"),
 		"-o", filepath.Join(tmp, "artifact.mender"),
@@ -259,17 +260,7 @@ func TestModifySigned(t *testing.T) {
 	err = run()
 	assert.NoError(t, err)
 
-	os.Args = []string{"mender-artifact", "modify",
-		"-n", "release-2",
-		filepath.Join(tmp, "artifact.mender")}
-
-	err = run()
-	assert.Error(t, err)
-	assert.Contains(t, err.Error(), ErrInvalidSignature.Error())
-
-	err = copyFile("mender_test.img", filepath.Join(tmp, "mender_test.img"))
-	assert.NoError(t, err)
-
+	// Modify the artifact and re-sign it using the same RSA private key.
 	os.Args = []string{"mender-artifact", "modify",
 		"-n", "release-2",
 		"-k", filepath.Join(tmp, "rsa.key"),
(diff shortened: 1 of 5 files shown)


cli/mender-artifact/modify_existing_test.go
```diff
@@ -252,42 +252,46 @@ func TestModifySigned(t *testing.T) {
...
-		"-k", filepath.Join(tmp, "rsa.key"),
+		"-n", "release-3",
 		filepath.Join(tmp, "artifact.mender")}
 
 	err = run()
 	assert.NoError(t, err)
 
+	// There is a surprising side-effect (probably a bug) in the modify command:
+	// it deletes the mender_test.img in the same directory.
+	// So for now, we have to copy it again before re-creating the artifact.
```
oleorhagen: That is worrying indeed, I actually think this might be a showstopper for this PR until it is fixed. Just so we don't have to merge code tests that are dependent upon a bug.

oleorhagen:
Apart from the comment above this is looking excellent! So clear that I actually had to check the unit test thrice just to make sure it just wasn't a modification of something I had written myself previously lol.

cli/mender-artifact/modify_existing_test.go
oleorhagen: I think that this might be related to a fix I did a few weeks back, where the update would change names when modifying an artifact. Thus when the functions clean up what they think are tmp-files returned from the unpacked artifact it is really deleting the original update, as they now share the same name.
https://github.com/mendersoftware/mender-artifact/commit/e28583185d184fecf110720ad62ef32721aee0aa#diff-0f63f0dd855c7e3c955b7fa7fe2a4577

oleorhagen:
Also I had to think about `Simplified the validate function so it no longer returns an
error if the artifact contains a signature but no public
key was specified by the caller.` a bit, but I think you are onto something here, and that this is actually a better way of doing it. @kacf thoughts?

oleorhagen:
https://github.com/mendersoftware/mender-artifact/pull/148

cli/mender-artifact/modify_existing_test.go
oleorhagen: This seems to do the trick. https://github.com/mendersoftware/mender-artifact/pull/148

cosinekitty: If you notice, I actually didn't change the test logic, I just added comments that explained the existing bug workaround, because it was confusing.  But yes, it would be nicer if your #148 could be merged first, then I can remove the bug workaround in #147. 

Also I created an issue for the bug:
https://tracker.mender.io/browse/MEN-2171

cosinekitty:
Yes, about that change to the validate function: I thought it might raise some eyebrows, but I can explain it this way. If an artifact is signed, that fact should not require the user of it to verify its signature. The user may or may not care that it is authentic, depending on the situation. 

I'm proposing this as a general principle, as a Mender user myself. The crux is that it is odd that a signed artifact should ever be less useful than an unsigned artifact.  In the case where we want to verify authenticity (such as during an update with a public key supplied), the unsigned artifact is failed just like a signed artifact with a bad signature: they are treated the same. But when the Mender client is configured with no public key for verifying the artifact, it would be weird to allow an unsigned artifact to pass through but reject the signed one!

So if there is no public key given, we should not fail an otherwise valid artifact, that is, an artifact whose checksum is valid. On the other hand, if a public key is supplied, then I as a user I am saying I do care, and that the artifact's digital signature must be present and must be proven to be authentic.

If there is a use case where we want `cat`, `cp`, etc., to fail on an unsigned (or worse, forged) artifact, then we would need to add support for the `-k` option on those commands too, so that the public key can be passed in. Even then, `-k' should be optional, and omitting it should not fail the command if the artifact is otherwise valid (checksum is correct).

cli/mender-artifact/modify_existing_test.go
cosinekitty: Gladly! I will do that later today. Thanks, guys...

cosinekitty: This PR should be all ready to merge now.

oleorhagen:
@cosinekitty Thank you so much for your contribution :)

PR #148, opened 2018-10-08
FIX: mender-artifact cat,cp,modify etc no longer removes the update.

oleorhagen:
Previously an update present in a directory, with the same name as the
update present in an update would be removed as a result of what the
functions thought was tmp-files.

Changelog: None


cosinekitty:
That fix looks good. I will test later today. By the way, I did create an issue for this:
https://tracker.mender.io/browse/MEN-2171

PR #150, opened 2018-10-30
[WIP] MEN-2224 Support uncompressed updates

mzimmermanngcx:
Just a preview to allow further discussions.

Currently the gocyclo test fails, so we probably have to refactor these functions:
```
16 main signExisting ./cli/mender-artifact/sign.go:27:1
16 main modifyArtifact ./cli/mender-artifact/modify.go:31:1
```

PR #156, opened 2018-11-24
Add support for compressing artifacts using LZMA

spencercw:
This depends on PR #150, which I have merged into this PR for the time being. Other than that I think it should be good to go. I don't have any hardware to hand so I can't test a real deployment, but the CLI tool can read the artifact back so I can't see any reason why it wouldn't work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ go:
 install:
     # Get tools used in Makefile
     - make get-tools
-    - sudo apt-get install mtools
+    - sudo apt-get install mtools liblzma-dev
 
 before_script:
     # Print build info.
--- a/LIC_FILES_CHKSUM.sha256
+++ b/LIC_FILES_CHKSUM.sha256
@@ -9,6 +9,7 @@ ceb1b36ff073bd13d9806d4615b931707768ca9023805620acc32dd1cfc2f680  vendor/github.
 2eb550be6801c1ea434feba53bf6d12e7c71c90253e0a9de4a4f46cf88b56477  vendor/github.com/pmezard/go-difflib/LICENSE
 2d36597f7117c38b006835ae7f537487207d8ec407aa9d9980794b2030cbc067  vendor/golang.org/x/sys/LICENSE
 2d36597f7117c38b006835ae7f537487207d8ec407aa9d9980794b2030cbc067  vendor/golang.org/x/crypto/LICENSE
+a284e538f3e5649ae6a8f32da2bbd18e9daabf2c1ebbfb99ed92e797b159be1a  vendor/github.com/mendersoftware/go-liblzma/LICENSE
 #
 # ISC licenses.
 3525392c6db3b804af76980b2c560ee9ec1abdadd907d76a26091df7c78f3a25  vendor/github.com/davecgh/go-spew/LICENSE
--- a/areader/reader_test.go
+++ b/areader/reader_test.go
@@ -312,10 +312,11 @@ func writeDataFile(t *testing.T, name, data string) io.Reader {
 	comp := artifact.NewCompressorGzip()
 
 	buf := bytes.NewBuffer(nil)
-	gz := comp.NewWriter(buf)
+	gz, err := comp.NewWriter(buf)
+	assert.NoError(t, err)
 	tw := tar.NewWriter(gz)
 	sw := artifact.NewTarWriterStream(tw)
-	err := sw.Write([]byte(data), name)
+	err = sw.Write([]byte(data), name)
 	assert.NoError(t, err)
 	err = tw.Close()
 	assert.NoError(t, err)
--- a/artifact/compressor.go
+++ b/artifact/compressor.go
@@ -16,32 +16,68 @@ package artifact
 
 import (
 	"io"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
 )
 
+var compressors = make(map[string]Compressor)
+
 type Compressor interface {
 	GetFileExtension() string
 	NewReader(r io.Reader) (io.ReadCloser, error)
-	NewWriter(w io.Writer) io.WriteCloser
+	NewWriter(w io.Writer) (io.WriteCloser, error)
+}
+
+func RegisterCompressor(id string, compressor Compressor) {
+	compressors[id] = compressor
 }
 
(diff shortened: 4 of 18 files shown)


spencercw:
Rebased on master with #150 applied and improved the test coverage a little.

kacf:
Isn't github.com/smira/lzma using the LZMA v1 algorithm? I'm not sure if this is a good addition to Mender, since the algorithm has been superseded by LZMA v2 (xz in particular), and we also have to think about binary size of the client.

spencercw:
Hi Kristian,

Re. LZMA v1, that's fair. I did look at github.com/ulikunitz/xz (which seems to be the 'popular' one), but it says on the GitHub page that it is under development and the API is not stable, so I passed that one by. There are other options available which I could look into.

Binary size is a bit of a trade off. In my use case, smaller update files are more important because updates will be deployed over a metered mobile data connection. Perhaps it could be added as a PACKAGECONFIG option to the mender-artifact recipe, so it can be turned off to reduce the binary size if it is unwanted?

kacf:
Yes, that would be acceptable, `no-lzma` tag for `go build` or something like that.

Another solution would be to pipe data to the system `xz` command instead of using a library, and have it perform the compression and decompression for us. This would give even more space savings, and I think `xz` is usually present in most configurations. But that means more implementation work, obviously.

spencercw:
I tend to prefer using libraries where possible, and I wouldn't expect there to be much size difference between invoking an executable and dynamically linking to the lzma library. The xz utility seems to dynamically link to liblzma, so:
- If xz is already installed, there is no significant size penalty since the lib is already there.
- If xz is not installed, then the size penalty would be for just liblzma, rather than xz+liblzma.

Either way, we're going to need to make sure it's available on the build server. By the looks of things I think I can just add it to the apt-get in travis.yml.

kacf:
Yes, if you can dynamically link to liblzma using a C binding, that would be even better.

spencercw:
So I've found two existing wrappers for liblzma: [github.com/danielrh/go-xz](https://github.com/danielrh/go-xz) and [github.com/remyoudompheng/go-liblzma](https://github.com/remyoudompheng/go-liblzma). Amazingly, they both have the exact same bug where if the source reader returns `nBytes > 0 && err == EOF` on the final successful read (instead of `nBytes == 0 && err == EOF` *after* the last successful read) then the decompressed output can be truncated.

I have raised a PR for the first, which as yet hasn't been acknowledged. There is already a PR open on the second, but it hasn't had any activity for some time.

On the whole, the code for the second seems a bit tidier and better tested so, this issue aside, I think that would be the one to use. Assuming there is no activity in the next week or so, I could fork it (or perhaps you would prefer to fork it under the mendersoftware account?) and merge the fix in there. What do you reckon?

kacf:
> So I've found two existing wrappers for liblzma: [github.com/danielrh/go-xz](https://github.com/danielrh/go-xz) and [github.com/remyoudompheng/go-liblzma](https://github.com/remyoudompheng/go-liblzma). Amazingly, they both have the exact same bug where if the source reader returns `nBytes > 0 && err == EOF` on the final successful read (instead of `nBytes == 0 && err == EOF` _after_ the last successful read) then the decompressed output can be truncated.

Ok, but does this actually happen in practice? I don't recall the exact behavior, but I would assume that `io.Reader` returns `nBytes == 10 && err == nil` if there are 10 bytes left, even if those are the last 10 bytes. And then it would return `err == EOF` afterwards. I could be wrong though.

> I have raised a PR for the first, which as yet hasn't been acknowledged. There is already a PR open on the second, but it hasn't had any activity for some time.
> 
> On the whole, the code for the second seems a bit tidier and better tested so, this issue aside, I think that would be the one to use. Assuming there is no activity in the next week or so, I could fork it (or perhaps you would prefer to fork it under the mendersoftware account?) and merge the fix in there. What do you reckon?

Assuming it's really needed (ref previous comment), merging the PR is obviously the most desirable, but a fork is ok too. Just let me know what the result is, and I can create the mendersoftware clone if needed.

spencercw:
> Ok, but does this actually happen in practice? I don't recall the exact behavior, but I would assume that io.Reader returns nBytes == 10 && err == nil if there are 10 bytes left, even if those are the last 10 bytes. And then it would return err == EOF afterwards. I could be wrong though.

Yes, the tar reader returns an 'early' EOF. Either way is acceptable according to the io.Reader documentation, which is just inviting bugs IMO. By coincidence, the test image I'm using happens to trigger the faulty behaviour and mender-artifact can't read back the artifact it has just written.

I'm not in any particular rush so I will just leave it a while to see if the PR gets merged.

kacf:
Alright, sounds good!

kacf:
It's done! We now have a new fork at https://github.com/mendersoftware/go-liblzma with your PR merged!

spencercw:
Nice one. Lets try this one on for size.

I tried uploading an lzma-compressed artifact to our hosted Mender account but unsurprisingly it got a bit upset. Not really sure how best to tackle that.
"Artifact couldn't be uploaded. reading artifact error: reader: found unexpected file in artifact: header.tar.xz... [Request ID: 22857036]"

kacf:
Yes, the deployments service would need to have its mender-artifact vendor library updated. There a couple of things that need to happen where, which I have described in [MEN-761](https://tracker.mender.io/browse/MEN-761). @spencercw, I don't see your name in the list of users there. Perhaps you want to create one?

dellgreen:
See https://tracker.mender.io/browse/MEN-761 for existing background work we have done on this which may help

spencercw:
Thanks, I wasn't aware of that other ticket. I'm out of the office for the rest of the year but I might be able to help with some of the outstanding changes when I get back.

oleorhagen:
@spencercw I merged this PR a bit pre-maturely, thinking that vendoring updates would not prove to be a problem. There is not a problem with the PR per-se, but the CGO dependency will take a little bit of work to get working with our deployments service. This should not be a lot of work, but too much for this last hour on the job, before the holidays. I promise that this will be the first thing I look into next year! 
Merry Christmas :)

spencercw:
Ok no worries, thanks for the update. Merry Christmas :)

spencercw:
FYI, the go-liblzma PR has now been merged upstream so we could drop the mendersoftware fork if you like.

PR #160, opened 2018-12-03
FIX: mender-artifact cat now cleans up resources on write-errors.

oleorhagen:
Changelog: Title


kacf:
Should be cherry-picked as well.

PR #161, opened 2018-12-20
MEN-761: Added a cgo tag to the lzma compressor.

oleorhagen:
As the lzma compressor is dependent upon go-liblzma, which has a cgo dependency,
this will not build when cgo is disabled.

Changelog: None


oleorhagen:
Failing tests are from another (old) test-run in parallell with another meta-menderPR.
This build: https://mender-jenkins.mender.io/job/mender-builder/2237/

PR #166, opened 2019-01-02
FIX: mender-artifact modify did not repack a sdimg after modification.

oleorhagen:
Changelog: Title

### DIFF
--- a/cli/mender-artifact/modify.go
+++ b/cli/mender-artifact/modify.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


oleorhagen:
2019 already :see_no_evil:

cli/mender-artifact/modify.go
```diff
@@ -61,7 +61,7 @@ func modifyArtifact(c *cli.Context) error {
 			defer os.Remove(mc.path)
 		}
 	} else if len(modifyCandidates) == 4 { // sdimg
-		modifyCandidates = modifyCandidates[1:2]
+		modifyCandidates = modifyCandidates[1:3]
```
kacf: I don't get it. Are we simply including rootfsB in the modification, and that's it? Correct me if I'm wrong, but this doesn't sound like it would affect vanilla images, which is the problem described in [MEN-2294](https://tracker.mender.io/browse/MEN-2294). Can you clarify?

oleorhagen:
The problem was that sdimgs would not be repacked, due to the test below, where it checks for whether a rootfsA and rootfsB is included for modification.

cli/mender-artifact/modify.go
oleorhagen: The problem was that the sdimg was not repacked due to the array length being too short, for the `if` repack test below. What do you mean by Vanilla-imags? Ticket says that only sdimg's are effected.

PR #168, opened 2019-01-16
Mender artifact v3: API adjustment + taxonomy fix

ghost:


kacf:
I think this one should be fairly straightforward, and it's blocking [MEN-2309](https://tracker.mender.io/browse/MEN-2309), so merging right away.

PR #170, opened 2019-01-28
Report a human readable error in case the artifact payload is not ext4. 

oleorhagen:
Changelog: Title


oleorhagen:
I originally planned for something fancier, like checking the filesystem type. However this turned out to be harder than i had thought. First of all, I created a small bfs filesystem, and created an Artifact out of it. This led to the expected failure, and obscure error message Mirza has encountered whilst using the tool. However - when I stat'ed the file that is being fsck, the stat_fs struct told me the magic byte was 0x53ef, which it clearly is not. All, in all, and long story short. This will do the trick I think. Kapisch?

PR #179, opened 2019-02-18
MEN-2331: Add rm functionality to mender-artifact.

oleorhagen:
Changelog: mender-artifact tool now supports removing a file in either an sdimg,
or in a Mender Artifact, with the rm command.


cli/mender-artifact/copy.go
```diff
@@ -37,7 +37,7 @@ func Cat(c *cli.Context) (err error) {
 	if !isimg.MatchString(c.Args().First()) {
 		return cli.NewExitError("The input image does not seem to be a valid image", 1)
 	}
-	r, err := NewPartitionFile(comp, c.Args().First(), c.String("key"))
+	r, err := virtualPartitionFile.Open(comp, c.Args().First(), c.String("key"))
```
oleorhagen: Changed the API so that it is looks more like a regular file (better semantics imo)

cli/mender-artifact/copy.go
```diff
@@ -96,10 +100,7 @@ func Copy(c *cli.Context) (err error) {
 	}
 
 	_, err = io.Copy(w, r)
-	// Ignore the close error in case of an io.Copy error.
-	if cerr := w.Close(); cerr != nil && err != nil {
-		err = cerr
-	}
+
```
lluiscampos: Just wondering.

So using defer w.Close will hide any error that could occur while actually closing, right? Is this a good practice? It seems like it is something we do a lot...

But for all these, can we instead do something like the following?

```
  defer func() {
    if ferr := f.Close(); ferr != nil {
      err = ferr
    }
  }()
```

oleorhagen: Yes, I think you are right here. This could potentially be misleading, as an error repacking, say an artifact, in the `Close()` method would indeed be camouflaged by this. Let me fix it up. Good eye!

lluiscampos: Keep in mind that this problem was already present in the old code. And it is probably present in other source files as well.

IMO, if you decide to fix it, it should be done everywhere to avoid inconsistency in the design. So please have a look around for suspicious `.close()` calls.

oleorhagen: Yes, of course :)

oleorhagen: @lluiscampos Now checks the error message of all the writers where it is necessary. Feel free to have a look.

oleorhagen:
@lluiscampos Now the close calls are checked whenever a write has been done. Looks better ?

lluiscampos:
@olehermanse Hi. Yes, it looks better.

It is in my to-do list for today to go through the rest of the code and approve the review. I'll be back to you later today :smiley:

oleorhagen:
haha, good :) Wrong Ole though. Let's leave @olehermanse alone

cli/mender-artifact/partition.go
```diff
@@ -341,8 +374,8 @@ func (f *fatFile) Write(b []byte) (n int, err error) {
 	if err = f.tmpf.Sync(); err != nil {
 		return 0, errors.Wrap(err, "fatFile: Write: Failed to sync tmpfile")
 	}
-	// Use MTools to write to the fat-partition
-	cmd := exec.Command("mcopy", "-i", f.path, f.tmpf.Name(), "::/"+f.imageFilePath)
+	f.tmpf.Close()
```
lluiscampos: Why do we `Close`, here? We do have the close already in `func (f *fatFile) Close() (err error)` method

cli/mender-artifact/partition.go
```diff
@@ -324,7 +357,7 @@ func newFatFile(imageFilePath string, partition partition) (*fatFile, error) {
 
 // Read Dump the file contents to stdout, and capture, using MTools' mtype
 func (f *fatFile) Read(b []byte) (n int, err error) {
-	cmd := exec.Command("mtype", "-i", f.path, "::/"+f.imageFilePath)
```
lluiscampos: Why did we need to change this?, isn't `imageFilePath` same as before? I see that in `Delete` we do trim the last `/`, but not in REad or Write. Is that intentional?

cli/mender-artifact/main.go
```diff
@@ -212,6 +212,20 @@ func run() error {
 		Action:      Copy,
 	}
 
+	remove := cli.Command{
+		Name:        "rm",
+		Usage:       "rm [artifact|sdimg|uefiimg]:<filepath>",
+		Description: "Removes the given file or directory from an Artifact or sdimg.",
```
lluiscampos: To be consistent, can we use `Artifact` also in the other help texts? copy, cat, install, ...

cli/mender-artifact/debugfs_test.go
```diff
@@ -42,7 +42,7 @@ func TestExecuteCommand(t *testing.T) {
 	}
 
 	for name, test := range tests {
-		err := executeCommand(test.cmd, "mender_test.img")
+		_, err := executeCommand(test.cmd, "mender_test.img")
```
lluiscampos: We could verify that `stdout` is actually `nill` here

cli/mender-artifact/debugfs.go
```diff
@@ -104,53 +105,100 @@ func debugfsReplaceFile(imageFile, newFile, image string) (err error) {
...
 	if err = cmd.Start(); err != nil {
-		return errors.Wrap(err, "debugfs: run debugfs script")
+		return nil, errors.Wrap(err, "debugfs: run debugfs script")
 	}
-	data, err := ioutil.ReadAll(ep)
-	if err != nil {
-		return err
+	if err := cmd.Wait(); err != nil {
+		return nil, err
 	}
```
lluiscampos: I think it is better to use `cmd.Run()` in this case instead of Start+Wait here, once there is no handling from our side of stderr/stdout

cli/mender-artifact/debugfs.go
```diff
@@ -104,53 +105,100 @@ func debugfsReplaceFile(imageFile, newFile, image string) (err error) {
...
+	cmd := fmt.Sprintf("cd %s", filepath.Dir(imageFile))
+	if _, err = executeCommand(cmd, image); err != nil {
+		return err
+	}
+	isDir := filepath.Dir(imageFile) == filepath.Clean(imageFile)
+	var deleteCmd string
+	if isDir {
+		return debugfsRemoveDir(imageFile, image, recursive)
+	} else {
+		deleteCmd = "rm"
```
lluiscampos: Maybe this should be out of the else

My reasoning: all of it should be in the else (set deleteCmd, set cmd, executeCommand) or none of it.

cli/mender-artifact/debugfs.go
```diff
@@ -104,53 +105,100 @@ func debugfsReplaceFile(imageFile, newFile, image string) (err error) {
...
+			regexp := regexp.MustCompile(`[0-9]+ +\([0-9]+\) +((.\w+.)+)`)
+			for _, filename := range regexp.FindAllStringSubmatch(buf.String(), -1) {
+				err = debugfsRemoveFileOrDir(filepath.Join(imageFile, string(filename[1])), image, recursive)
+				if err != nil {
+					return errors.Wrap(err, "debugfsRemoveDir")
+				}
+			}
+			// Now that all the subdirs and files should be removed, try and remove the directory
+			// once more.
+			return debugfsRemoveDir(imageFile, image, recursive)
```
lluiscampos: Shouldn't this last call be with recursive=false? As the comment says, in theory there is no more files or subdirs, so we shouldn't need to recursive here in any case.

cli/mender-artifact/copy_test.go
```diff
@@ -387,6 +387,54 @@ func TestCopy(t *testing.T) {
...
+					imgpath + ":/boot/test.txt"}
+				err := run()
+				assert.Error(t, err)
+			},
+		},
+		{
+			name:        "Delete a directory from a fat partition",
+			validimages: []int{2}, // only valid for fat boot partition
+			argv:        []string{"mender-artifact", "rm", "-r", ":/uboot/testdir/"},
+		},
```
lluiscampos: There is no specific test for ext partition (i.e. `validimages: {1}`). I guess it is already covered in general rm tests (the ones without specifying `validimages`)?.

cli/mender-artifact/copy.go
lluiscampos: Not an expert in golang (yet!), but I thought that any symbol to be exported in a file needs to start with capital letter. I would then expect the constant `virtualPartitioFile` to be `VirtualPartitionFile`.

I also wonder about the reason for such design. I would have expected a NewPartitionFile method returning an object that satisfy VPFile interface, and from there the user can call open/close, etc

Thanks for clarifying!

cli/mender-artifact/copy.go
```diff
@@ -57,34 +57,44 @@ func Copy(c *cli.Context) (err error) {
...
 		if err != nil {
-			w.Close()
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
 		}
+		w = vfile
 	case copyout:
-		r, err = NewPartitionFile(comp, c.Args().First(), c.String("key"))
-		defer r.Close()
+		vfile, err = virtualPartitionFile.Open(comp, c.Args().First(), c.String("key"))
+		defer vfile.Close()
```
lluiscampos: `defer wclose`

lluiscampos: Other places out of Copy and Delete are missing this improvement, but it is not a big deal so I won't insist more. Only this one I insist for the sake of consistency inside the same function :smiley:

cli/mender-artifact/main.go
oleorhagen: Changed it up for everything except cat, where the usage text has it covered: `Copy from or into an artifact, or sdimg where either the <src> or <dst> has to be of the form [artifact|sdimg]:<filepath>`
Otherwise the usage line will be very long.

PR #181, opened 2019-03-04
Updated code-coverage in rootfs_image.go

oleorhagen:

### DIFF
--- a/artifact/tar_writer_stream.go
+++ b/artifact/tar_writer_stream.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -43,6 +43,9 @@ func NewTarWriterStream(w *tar.Writer) *StreamArchiver {
 }
 
 func (str *StreamArchiver) Write(data []byte, archivePath string) error {
+	if str.Writer == nil {
+		return errors.New("arch: Can not write to empty tar-writer")
+	}
 	hdr := &tar.Header{
 		Name: archivePath,
 		Mode: 0600,
--- a/handlers/common.go
+++ b/handlers/common.go
@@ -184,6 +184,9 @@ func match(pattern, name string) bool {
 }
 
 func writeFiles(tw *tar.Writer, updFiles []string, dir string) error {
+	if tw == nil {
+		return errors.New("writer: tar-writer is nil")
+	}
 	files := new(artifact.Files)
 	for _, u := range updFiles {
 		files.FileList = append(files.FileList, u)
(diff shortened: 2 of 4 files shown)


artifact/tar_writer_stream.go
```diff
@@ -43,6 +43,9 @@ func NewTarWriterStream(w *tar.Writer) *StreamArchiver {
 }
 
 func (str *StreamArchiver) Write(data []byte, archivePath string) error {
+	if str.Writer == nil {
+		return errors.New("arch: Can not write to empty tar-writer")
```
oleorhagen: I think it is better to return an error than to panic on an empty tar writer.

handlers/common.go
```diff
@@ -184,6 +184,9 @@ func match(pattern, name string) bool {
 }
 
 func writeFiles(tw *tar.Writer, updFiles []string, dir string) error {
+	if tw == nil {
```
oleorhagen: same

handlers/rootfs_image_test.go
```diff
@@ -148,3 +150,183 @@ func TestRootfsReadHeader(t *testing.T) {
...
+			rfs:  NewRootfsV2(""),
+			args: ComposeHeaderArgs{},
+			err:  errors.New("writer: tar-writer is nil"),
+		},
+		"version 3 - no tar writer": {
+			rfs:  NewRootfsV3(""),
+			args: ComposeHeaderArgs{},
+			err:  errors.New("ComposeHeader: Payload: can not tar type-info: arch: Can not write to empty tar-writer"),
+		},
+		"version 1 - success": { // TODO - should this succeed with no update files?
```
oleorhagen: Not sure. Which is correct?

handlers/rootfs_image_test.go
```diff
@@ -148,3 +150,183 @@ func TestRootfsReadHeader(t *testing.T) {
...
+			if test.verifyFunc != nil {
+				test.verifyFunc(test.args)
+			}
+		}
+	}
+}
+
+func TestNewAugmentedInstance(t *testing.T) {
+	rfs := NewRootfsInstaller()
+	inst, err := rfs.NewAugmentedInstance(NewRootfsV2(""))
```
oleorhagen: The rest are tests for the setters and getters of the udpates. A fair bit of code. Not really sure how helpful they are though, except increasing the coverage. It did find the nil pointer tar writer panic though.

handlers/rootfs_image_test.go
kacf: I'm not sure either, but I wouldn't worry too much about it. We have completely disabled v1 support on the reading side, in anticipation of removing it altogether.

handlers/rootfs_image_test.go
kacf: Perhaps not useful right now, but they will be in the future, when we switch rootfs updates over to augmented updates (which was the original plan). For now it's a future investment.

artifact/tar_writer_stream.go
kacf: :+1:

handlers/common.go
kacf: :+1:

PR #183, opened 2019-03-13
Apply coverage analysis to all subpkgs using -coverpkg flag

lluiscampos:
Changelog: None


lluiscampos:
@kacf Please have a look.

If you like this solution I will then port it to `mender` both master and update_modules, so that we can compare the coverage

PR #184, opened 2019-03-13
Use GetBinaryPath helper to search PATH and sbin

andhe:
Certain commands are spawned from the mender-artifact command and those
commands may sometimes live outside of $PATH (since /sbin, /usr/sbin,
.. is normally not in a regular users path). First search PATH but
also try other locations.

Adresses: https://tracker.mender.io/browse/MEN-2180
Changelog: fixes issue when binary dependencies are not in PATH
Signed-off-by: Andreas Henriksson <andreas@fatal.se>

### DIFF
--- a/utils/binpath.go
+++ b/utils/binpath.go
@@ -1,0 +1,40 @@
+// Copyright 2019 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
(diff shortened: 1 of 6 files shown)


utils/binpath.go
```diff
@@ -0,0 +1,40 @@
...
+
+package utils
+
+import (
+	"os/exec"
+	"path"
+)
+
+func GetBinaryPath(command string) string {
+	// first check if debugfs exists in PATH
```
lluiscampos: first check if *command* exists in PATH

PR #187, opened 2019-03-26
Add docker container image

ryansch:
This adds a Dockerfile for running `mender-artifact` anywhere that docker runs.  I've tested it successfully on macOS.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+# Docker build uses published asset, no code needed
+*
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM buildpack-deps:stretch-curl as tini
+
+ENV TINI_VERSION v0.18.0
+
+RUN mkdir -p /tmp/build && \
+    cd /tmp/build && \
+    wget -O tini https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini && \
+    wget -O tini.asc https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini.asc && \
+    gpg --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 595E85A6B1B4779EA4DAAEC70B588DFF0527A9B7 && \
+    gpg --verify tini.asc && \
+    cp tini /sbin/ && \
+    chmod +x /sbin/tini && \
+    cd /tmp && \
+    rm -rf /tmp/build && \
+    rm -rf /root/.gnupg
+
+FROM debian:latest
+
+COPY --from=tini /sbin/tini /sbin/
+
+ENV MENDER_ARTIFACT_VERSION 2.4.0
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        wget \
+        ca-certificates \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/* \
+    && cd /usr/bin \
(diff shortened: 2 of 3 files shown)


kacf:
Thanks!

@estenberg: What do you think about this? I'm thinking that we should rather focus on getting real Mac and Windows binaries, like we discussed, especially since this approach is not entirely straightforward, neither from a maintenance perspective (versioning especially), nor the usage perspective. Unlike mender-convert, mender-artifact doesn't need a lot of privileges or dependencies.

estenberg:
Thank you for contributing @ryansch !

I would tend to agree with @kacf, mender-artifact is such a simple tool that it is probably better to just provide a native Mac version.

What do you think @ryansch  -- did you try compiling it natively for Mac?

mzedel:
Thanks @ryansch, I for one would also like to just run containers everywhere - even if it's just to not have to worry about dependencies/ compatibilities.
Besides the discussion about a native compilation option, there is however one thing that I would suggest based on the PR itself. Try to base the final stage on something smaller (even if debian:latest is not too shabby with the current 45mb) especially if it should be redistributed at some point.

Incidentally I also worked on a contained build of mender-artifact which is basically what's in https://github.com/mendersoftware/mender-artifact/pull/188 and would at least get rid of the versioning problem as it is built straight from the repo content. Maybe you want to have a look at that - I admittedly have not tested creating artifacts with it, but it *should* work similar to your approach.

But as @estenberg and @kacf suggested a native compilation also be possible without a lot of problems.

ryansch:
@mzedel If you're going as small as possible you can do a static golang build and use a `FROM scratch` container.  I was trying to get something useful out the door without digging into the build specifics.

ryansch:
@estenberg I did not.  I'm pretty comfortable with running linux binaries in a docker container as many other non-golang projects function that way.  I'd love a native binary though.

estenberg:
OK, makes sense. It will take a little while until we are distributing a native MacOS mender-artifact binary though. At that point I don't think this container will be that useful and will probably not be maintained.

@kacf How about moving this into a support/ directory or something similar, indicating this is not officially part of mender-artifact (and may be removed/untested)? Similar to what we have for mender: https://github.com/mendersoftware/mender/tree/master/support

ryansch:
@estenberg Can I suggest that you build and push the image as part of your deploy process for new artifacts?

estenberg:
Do you mean release this docker image as officially supported together with Mender releases?

I think we'd have to discuss a bit the right path forward here, especially given that a native Mac version is planned and it would supersede this. If we introduce official support for something then people will start using it and deprecation / removal will be painful for many.

Does that make sense, or did I misunderstand?

ryansch:
I see where you're coming from.  My hope would be for you to support both native binaries and a docker image but I can understand where you're coming from.  Honestly, I'll just treat the docker image as best effort.

kacf:
> How about moving this into a support/ directory or something similar, indicating this is not officially part of mender-artifact (and may be removed/untested)? Similar to what we have for mender:

Since we are actively pursuing a different solution, I think a better approach is simply for @ryansch to maintain this in his own fork. Much easier for both parties, since we are not planning on maintaining this anyway, and he won't have to go through us.

So I'm going to close this, but I encourage people to visit [@ryansch's fork](https://github.com/outstand/mender-artifact/tree/add-docker-image) and clone his docker branch, if they want Mac support through a Docker image:

```
git clone -b add-docker-image https://github.com/outstand/mender-artifact.git
```

ryansch:
@kacf I'll instead point users at https://github.com/outstand/docker-mender-artifact which is a dedicated repository for maintaining the docker container image.

kacf:
:+1:

PR #188, opened 2019-03-27
Added multi stage dockerfile to contain builds

mzedel:
While moving other parts of mender to contained docker builds, mender-artifact would also be a nice to have. Due to the lzma dependency this is not based on scratch or busybox, but alpine.

Changelog: None
Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,12 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+build-contained:
+	rm -f mender-artifact; \
+	image_id=$$(docker build -f Dockerfile . | awk '/Successfully built/{print $$NF;}'); \
+	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact /binary"; \
+	docker image rm $$image_id;
+
 install:
 	cd $(INSTALL_DIR) && $(GO) install $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ TOOLS = \
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
 GO_LDFLAGS = \
-	-ldflags "-X main.Version=$(VERSION)"
+	-ldflags "-extldflags '-static' -X main.Version=$(VERSION)"
 
 ifeq ($(V),1)
 BUILDV = -v
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ TOOLS = \
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
 GO_LDFLAGS = \
-	-ldflags "-extldflags '-static' -X main.Version=$(VERSION)"
+	-ldflags "-X main.Version=$(VERSION)"
 
 ifeq ($(V),1)
 BUILDV = -v
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile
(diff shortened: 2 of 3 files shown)


kacf:
The Travis error is being fixed in #189.

Makefile
```diff
@@ -24,7 +24,7 @@ TOOLS = \
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
 GO_LDFLAGS = \
-	-ldflags "-extldflags '-static' -X main.Version=$(VERSION)"
+	-ldflags "-X main.Version=$(VERSION)"
```
kacf: Out of curiosity: Why did you remove this?

I saw the warnings when this was readded, but it worked for me, and this could be quite desirable, especially on Mac and Windows, where the the lzma library is less likely to exist.

mzedel: here I just removed it to make as minimal a change as possible to the way it works now, as I didn't want to cause a breaking change somewhere else. It worked in the container as well but I wasn't sure about the warnings...

mzedel:
@kacf to be able to extract the binary afterwards I had to use a base image with a shell (just the `sh`) (another reason not to build this from `scratch`) - then I can run the image with the shell instead and mount a volume where the binary should go, and copy the binary out of the container. And since the shell has nothing else to execute, the shell and with it the container shuts down as soon as the copy finishes.
This way the container should be able to be run almost like an executable and we have a unified container for build artifacts...

Makefile
```diff
@@ -42,6 +42,16 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+build-contained:
+	if [ -f mender-artifact ]; \
+	then \
+		echo "mender-artifact binary already exists, please move or rename"; \
```
lluiscampos: Why don't we just remove unconditionally mender-artifact binary?

Makefile
```diff
@@ -42,6 +42,16 @@ endif
...
+		exit 1; \
+	fi; \
+	image_id=$$(docker build -f Dockerfile . | awk '/Successfully built/{print $$NF;}'); \
```
lluiscampos: For getting image_id, isn't in easier to `--tag` the build to a known name?

Makefile
```diff
@@ -42,6 +42,16 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+build-contained:
```
lluiscampos: It is a nice addition to our Makefile ;)

Makefile
mzedel: that might overwrite an existing tag and I wanted to avoid that... - will still remove an existing image with the same hash

mzedel:
@kacf now I finally got around the make differences and added the contained build...

Makefile
mzedel: good question... would be in line with the regular `build` task...

lluiscampos:
The test failure can be ignored, it is this spurious server error that we see some times.

As @kacf is on vacation, should I approve this?

Replace the if/else for an unconditional remove of the mender-artifact in the `build-contained` target of the `Makefile` and I'll be happy to do so.

Makefile
```diff
@@ -42,6 +42,15 @@ endif
...
+	if [ -f mender-artifact ]; \
+	then \
+		rm mender-artifact; \
+	fi; \
```
lluiscampos: You can simplify this even more by just running `rm -f mender-artifact`.

IMO, the least conditionals and for loops in Makefiles, the better. Only the strictly needed ones.

lluiscampos:
I dropped a new comment, sorry. Other than that looks good and I am approving it.

I don't think it is important to wait for the tests, as we had a pass last week. So feel free to merge it if it is blocking you.

Makefile
mzedel: my bad, I just thought of either regular `rm` (which potentially results in a warning) or reusing the condition... is `rm -f` now

lluiscampos:
Last commit is not signed.

Also, in the client team we try to group the changes in as little commits as possible. Rule of thumb is to rebase commits that amend previous commits (i.e. squash them), and in general keep the change set consistent with its log message.

Can you please tidy up a bit?

Thanks

mzedel:
@lluiscampos now I squashed the last commits and fixed the signoff. Shall I squash some more?

lluiscampos:
It is good :+1: 

Ignore the build errors and merge.

PR #189, opened 2019-03-28
Remove unit testing with go version 1.10

lluiscampos:
Changelog: None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: go
 
 # Golang version matrix
 go:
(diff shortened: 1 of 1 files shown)


lluiscampos:
Actual change:

Remove unit testing with go version 1.10
    
It is causing problems at the moment due to a change in upstream. And
generally speaking, we would like to support only one go version and to
be consistent across the mender components (yocto build, mender client,
mender-artifact...)
    
Changelog: None
    

PR #190, opened 2019-04-15
Contained cross compilation

mzedel:
so far this seems to compile and work on mac and partly on windows, but on windows it fails to access artifacts with:
`reader: error getting header update number: can not get update order from tar path`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,6 +42,22 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+PLATFORMS := darwin linux windows
+
+GO_LDFLAGS_WIN = -ldflags "-X main.Version=$(VERSION) -linkmode=internal -s -w -extldflags '-static' -extld=x86_64-w64-mingw32-gcc"
+
+buildNatives:
+	@arch="amd64";
+	@echo "building mac";
+	@env GOOS=darwin GOARCH=$$arch CGO_ENABLED=0 \
+		$(GO) build -a $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME)-darwin $(BUILDFILES) ;
+	@echo "building linux";
+	@env GOOS=linux GOARCH=$$arch \
+		$(GO) build -a $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME)-linux $(BUILDFILES) ;
+	@echo "building windows";
+	@env GOOS=windows GOARCH=$$arch CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ \
+		$(GO) build $(GO_LDFLAGS_WIN) $(BUILDV) -tags $(TAGS) nolzma -o $(PKGNAME)-windows.exe $(BUILDFILES) ;
+
 build-contained:
 	rm -f mender-artifact; \
 	image_id=$$(docker build -f Dockerfile . | awk '/Successfully built/{print $$NF;}'); \
--- a/Dockerfile.binaries
+++ b/Dockerfile.binaries
@@ -1,0 +1,16 @@
+FROM golang:1.11 as builder
+RUN apt-get update && \
+    apt-get install -y \
+        gcc gcc-mingw-w64 gcc-multilib \
+        git make \
+        musl-dev liblzma-dev
+RUN mkdir -p /go/src/github.com/mendersoftware/mender-artifact
+WORKDIR /go/src/github.com/mendersoftware/mender-artifact
+RUN go get 'golang.org/x/sys/windows'
+ADD ./ .
+RUN make buildNatives
+
+FROM alpine:3.9
+RUN apk add xz-dev
+COPY --from=builder /go/src/github.com/mendersoftware/mender-artifact/ /go/bin/
+ENTRYPOINT [ "/go/bin/mender-artifact-linux" ]
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -30,7 +30,6 @@ import (
 	"github.com/mendersoftware/mender-artifact/utils"
 
 	"github.com/pkg/errors"
-	"github.com/urfave/cli"
 )
 
 type writeUpdateStorer struct {
@@ -290,7 +289,7 @@ func repack(comp artifact.Compressor, artifactName string, from io.Reader, to io
 	return ar, err
 }
 
-func repackArtifact(comp artifact.Compressor, artifact, rootfs, key, newName string) error {
+func repackArtifact(comp artifact.Compressor, artifact, rootfs, newName string) error {
 	art, err := os.Open(artifact)
 	if err != nil {
 		return err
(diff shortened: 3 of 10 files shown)


Makefile
```diff
@@ -42,6 +42,27 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+PLATFORMS := darwin linux windows
+
+WIN_FLAGS = -ldflags "-X main.Version=$(VERSION) -linkmode=internal -s -w -extldflags '-static' -extld=x86_64-w64-mingw32-gcc"
```
oleorhagen: Why both internal and external linkflags?

Makefile
```diff
@@ -42,6 +42,27 @@ endif
...
+
+buildNatives:
+	@for platform in $(PLATFORMS); \
+	do \
+		arch="amd64"; \
+		output_name="$(PKGNAME)-$$platform"; \
+		if [ $$platform = windows ]; then \
+			echo $$output_name; \
+			echo "asdasddassad"; \
```
oleorhagen: Mean only for testing mayhaps?

Makefile
```diff
@@ -42,6 +42,27 @@ endif
...
+			wintags='$(TAGS) nolzma'; \
```
oleorhagen: Never lzma support on windows?

Makefile
```diff
@@ -42,6 +42,27 @@ endif
...
+			env CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ GOOS=$$platform GOARCH=$$arch $(GO) build $(WIN_FLAGS) -tags $$wintags $(BUILDV) $(BUILDFILES) ; \
+		else \
```
oleorhagen: Keywords can have newlines

Dockerfile
```diff
@@ -1,11 +1,12 @@
...
 ADD ./ .
-RUN make build
+RUN make buildNatives
 
 FROM alpine:3.9
 RUN apk add xz-dev
-COPY --from=builder /go/src/github.com/mendersoftware/mender-artifact/mender-artifact /go/bin/mender-artifact
-ENTRYPOINT [ "/go/bin/mender-artifact" ]
+COPY --from=builder /go/src/github.com/mendersoftware/mender-artifact/ /go/bin/
+ENTRYPOINT [ "/go/bin/mender-artifact-linux" ]
```
oleorhagen: Are they called this on all three platforms?

Makefile
mzedel: my best guess would be staying close to the regular linkflags (aka copy & paste remnant) while being able to compile a windows binary

Makefile
mzedel: argh, I though I made that afterwards should be "gone in 60 seconds"

Makefile
mzedel: that was the go to solution, as the compiler complained about not being able to find the `lzma.h` and the exploratory nature

Dockerfile
mzedel: the entrypoint is only a placeholder here and will be overridden during binary extraction, as the image will contain all builds (so this probably shouldn't be widely used)

but from all 3 binaries, the linux on would be the one able to be run in this container...

apart from that the name always is "mender-artifact-{linux|mac|windows.exe}"

Dockerfile
```diff
@@ -1,11 +1,12 @@
-FROM golang:1.11-alpine3.9 as builder
-RUN apk update &&  apk add gcc musl-dev libc6-compat git make xz-dev
+FROM golang:1.11 as builder
+RUN apt-get update && apt-get install -y gcc gcc-mingw-w64 gcc-multilib git musl-dev libc6 liblzma-dev make
```
oleorhagen: Maybe we should multiline this for prettiness?

oleorhagen:
Also the error on windows is a bit worrying :/

mzedel:
@oleorhagen I hope the switch to a separate Dockerfile clarifies the intention + thanks to the alignment in the makefile I noticed, that the `validate.exe` apparently was the build result, if no output name was set and the present `.exe` was a remainder of previous builds, carried over to subsequent builds the ADD in the Dockerfile - and new timestamps due to the multi stage nature lead me to believe the file was the result of the compilation... so thanks a lot for your review!

oleorhagen:
Great. Yes, the intention is more clear now. As long as this is contained in your build pipeline, I don't see no reason to dig deeper into this. 
Looks good :)

Makefile
```diff
@@ -42,6 +42,33 @@ endif
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(PKGNAME) $(BUILDFILES)
 
+PLATFORMS := darwin linux windows
+
+WIN_FLAGS = -ldflags "-X main.Version=$(VERSION) -linkmode=internal -s -w -extldflags '-static' -extld=x86_64-w64-mingw32-gcc"
```
kacf: For consistency, I would call this `GO_LDFLAGS_WIN` (more common to put platform at the end).

Makefile
```diff
@@ -42,6 +42,33 @@ endif
...
+		arch="amd64"; \
+		output_name="$(PKGNAME)-$$platform"; \
+		if [ $$platform = windows ]; then \
+			echo $$output_name; \
+			echo "asdasddassad"; \
+			wintags='$(TAGS) nolzma'; \
+			env CGO_ENABLED=1 CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ GOOS=$$platform GOARCH=$$arch $(GO) build $(WIN_FLAGS) -tags $$wintags $(BUILDV) $(BUILDFILES) ; \
+		else \
+			echo $$output_name; \
+			env CGO_ENABLED=0 GOOS=$$platform GOARCH=$$arch $(GO) build -a $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $$output_name $(BUILDFILES) ; \
```
kacf: Why `CGO_ENABLED=0`? Lzma requires it, no?

Should we just lose `CGO_ENABLED` from both of these lines? It's already set globally at the top.

Makefile
```diff
@@ -24,7 +24,7 @@ TOOLS = \
 VERSION = $(shell git describe --tags --dirty --exact-match 2>/dev/null || git rev-parse --short HEAD)
 
 GO_LDFLAGS = \
-	-ldflags "-X main.Version=$(VERSION)"
+	-ldflags "-X main.Version=$(VERSION) -linkmode=internal -s -w"
```
kacf: Can you explain why this is needed? I think this won't work on Yocto, because Yocto needs a standard ELF binary with a symbol table. There are ways around it, but let's make sure it's needed first.

Makefile
mzedel: without the CGO_ENABLED=0 the darwin build will crash

as for the lzma requirement: my guess was that if it runs in the container, it should be able to be run on linux... is there a specific functionality that I should try to run in order to see if it works with lzma?

Makefile
mzedel: Ah you're right, that seems to be left from before I moved these to the separate windows flags, thanks! I removed the corresponding part...

Makefile
oleorhagen: In the tests?
Locally you can just run mender artifact, and then try to create an artifact with the zlma (xz-extension) option for the artifact.

kacf: > as for the lzma requirement: my guess was that if it runs in the container, it should be able to be run on linux... is there a specific functionality that I should try to run in order to see if it works with lzma?

Yes, try this:

```
./mender-artifact --compression lzma write rootfs-image -n myname -t mydevice -f /dev/null
```

It's ok if this doesn't work on Mac, but it should work on Linux, and I suspect that `CGO_ENABLED=0` will prevent that.

Makefile
kacf: It's the makefile that needs the backslash, not the shell.

mzedel:
@kacf although I squashed and rebased it, there seems to be some problems with the commits

kacf:
Try this command:

```
git rebase HEAD~2 --onto origin/master
```

mzedel:
closing in favour of https://github.com/mendersoftware/mender-artifact/pull/198

PR #191, opened 2019-04-20
Switch to original go liblzma vendor

lluiscampos:
Switch to original source for vendored library go-liblzma

It seems like the PR we were interested it having merged has been merged
upstream, so there is no more interest in keeping our fork alive.

For creating .deb packages for the official Debian repositories, we are
encouraged to use the original source as a dependency. Hence, in order
to be consistent, it is cleaner to use the original one also in the
vendored folder.

Changelog: None


lluiscampos:
@kacf What do you think about this?

PR #185, opened 2019-03-22
checking if fsck is on path and returning error if not.

merlin-northern:
@kacf please take a look. tested:

`#./mender-artifact modify /tmp/qemux86-64-uefi-grub_release_1_master.mender --server-uri "$MENDER_SERVER_URL" --tenant-token "$MENDER_TENANT_TOKEN"
Error modifying artifact[/var/folders/c_/j5f_xrnj3790c0pbm3bg_rwm0000gn/T/core-image-full-cmdline-qemux86-64.ext4]: fsck.ext4 command not found.
`

PR #192, opened 2019-04-20
Look for fsck.ext4 via GetBinaryPath

lluiscampos:
Same as for the rest of binaries expected to be in /sbin or /usr/sbin.
It seems like this one was missed in commit 91d0f27 (PR #184)

Changelog: None


lluiscampos:
This PR tries to amend a change missing from @andhe in PR #184 and fixed with a different approach by @merlin-northern  on PR #185. For consistency, I'd like to use the util function when applicable.

@kacf Please have a look

PR #193, opened 2019-04-24
MEN-2488: Fix name modify command for rootfs-image Artifacts

lluiscampos:
It was not working for Artifact V3 headers. Extended the unit tests to
cover both supported versions.

Changelog: Title

Signed-off-by: Lluis Campos <lluis.campos@northern.tech>

PR #194, opened 2019-04-25
MEN-2486: Unsupport signed modify

lluiscampos:
MEN-2486: Remove support for re-signing modified artifacts

It was not properly implemented; the "-k" parameter was being used as
the public key to verify the signature of the current Artifact and as
the private key as well to sign the modifyied Artifact.

In any case, it is better to completely remove this option and use the
mender-convert sign after modifications to produce the final Artifact

Removed the functionality and improved the test to actually check
something after the modifications.

Changelog: `mender-artifact modify` does not support anymore signing the
Artifact after the modification. Use `mender-convert sign` after the
modification to sign the Artifact.

### DIFF
--- a/cli/mender-artifact/modify.go
+++ b/cli/mender-artifact/modify.go
@@ -42,7 +42,7 @@ func modifyArtifact(c *cli.Context) error {
 	}
 
 	modifyCandidates, isArtifact, err :=
-		getCandidatesForModify(c.Args().First(), nil)
+		getCandidatesForModify(c.Args().First())
 
 	if err != nil {
 		return cli.NewExitError("Error selecting images for modification: "+err.Error(), 1)
@@ -75,7 +75,7 @@ func modifyArtifact(c *cli.Context) error {
 	if isArtifact {
 		// re-create the artifact
 		err := repackArtifact(comp, c.Args().First(), modifyCandidates[0].path,
-			"", c.String("name"))
+			c.String("name"))
 		if err != nil {
 			return cli.NewExitError("Can not recreate artifact: "+err.Error(), 1)
 		}
--- a/cli/mender-artifact/modify.go
+++ b/cli/mender-artifact/modify.go
@@ -15,9 +15,7 @@
 package main
 
 import (
-	"bytes"
 	"encoding/json"
-	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -43,13 +41,8 @@ func modifyArtifact(c *cli.Context) error {
 		return cli.NewExitError("File ["+c.Args().First()+"] does not exist.", 1)
 	}
 
-	pubKey, err := processModifyKey(c.String("key"))
-	if err != nil {
-		return cli.NewExitError("Error processing private key: "+err.Error(), 1)
-	}
-
 	modifyCandidates, isArtifact, err :=
-		getCandidatesForModify(c.Args().First(), pubKey)
+		getCandidatesForModify(c.Args().First(), nil)
 
 	if err != nil {
 		return cli.NewExitError("Error selecting images for modification: "+err.Error(), 1)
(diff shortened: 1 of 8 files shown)


lluiscampos:
Based on PR #193, as we need the fix here also.

So here focus only on commit 9a377d5

lluiscampos:
Commit 892d702 belongs also to this PR

lluiscampos:
And commit bfafac8

Sorry but new small issues keep popping up and I'd like to fix them

cli/mender-artifact/modify.go
```diff
@@ -43,13 +41,8 @@ func modifyArtifact(c *cli.Context) error {
...
 	}
 
-	pubKey, err := processModifyKey(c.String("key"))
-	if err != nil {
-		return cli.NewExitError("Error processing private key: "+err.Error(), 1)
-	}
-
 	modifyCandidates, isArtifact, err :=
-		getCandidatesForModify(c.Args().First(), pubKey)
+		getCandidatesForModify(c.Args().First(), nil)
```
kacf: It looks like all calls to this function pass either `nil` or empty buffer now. Remove the argument?

cli/mender-artifact/modify.go
```diff
@@ -82,7 +75,7 @@ func modifyArtifact(c *cli.Context) error {
 	if isArtifact {
 		// re-create the artifact
 		err := repackArtifact(comp, c.Args().First(), modifyCandidates[0].path,
-			c.String("key"), c.String("name"))
+			"", c.String("name"))
```
kacf: There is only one other instance of this function call too, in `partition.go`. I *think* it is being used during `cp` stage. Have you checked the behavior of signing key there? Maybe it should be removed from both places, and the function argument `key` removed, like the previous one.

cli/mender-artifact/modify_existing_test.go
```diff
@@ -287,45 +287,44 @@ func TestModifySigned(t *testing.T) {
...
-		"-n", "release-1", "-f", filepath.Join(tmp, "mender_test.img"),
-		"-o", filepath.Join(tmp, "artifact_ecdsa.mender"),
-		"-k", filepath.Join(tmp, "ecdsa.key")}
-	err = run()
-	assert.NoError(t, err)
+		go func() {
+			err = run()
+			assert.NoError(t, err)
+			w.Close()
+		}()
```
kacf: Confused. You want to run this in the background? Doesn't it then return immediately, regardless of the result?

lluiscampos: I took this from other test that also tried to get the Stdout output.

I think that we run it in the background so that we can call ioutilReadAll(r), which will block until we call w.Close() from the goroutine.

Is there a simpler way to capture the Stdout here?

cli/mender-artifact/modify.go
lluiscampos: :+1:

cli/mender-artifact/modify.go
lluiscampos: Thanks, I will look into it in detail and see if we can remove it

cli/mender-artifact/modify.go
lluiscampos: Well, it is also used by `virtualPartitionFile.Open`, which is used for cp, cat, install and rm. So it falls into the same category as you other comment in `repackArtifact`: to verify the actual usage of the key in these operations and decide.

lluiscampos: Fixed. See last commit.

cli/mender-artifact/modify.go
lluiscampos: Fixed. See last commit

cli/mender-artifact/modify_existing_test.go
kacf: No, you're right. I missed that part.

lluiscampos:
@kacf As we have feared, the failure in `test_update_image_recovery` of the previous run, has not been reproduced in the the latest fun of Jenkins. I also run the test locally 10 times (using this version of `mender-artifact`) and still was not able to reproduce.

Furthermore, I cannot find any theoretical reason for my changes in mender-artifact to affect this test at all.

Merging.

PR #196, opened 2019-04-27
MEN-2487: Not supported module-image operations

lluiscampos:
Contains also changes from PR #194.

For this PR, the commits under review are from 38ae1c5 onward.

### DIFF
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -158,10 +158,16 @@ func unpackArtifact(name string) (string, error) {
 		return "", errors.Wrap(err, "failed to register install handler")
 	}
 
-	err = aReader.ReadArtifact()
+	r, err := read(aReader, nil, nil)
 	if err != nil {
 		return "", err
 	}
+
+	inst := r.GetHandlers()
+	if ! (len(inst) == 1 && inst[0].GetUpdateType() == "rootfs-image" ) {
+		return "", errors.New("Only rootfs update types supported")
+	}
+
 	// Give the tempfile it's original name, so that the update does not change name upon a write.
 	tmpfilePath := tmp.Name()
 	newNamePath := filepath.Join(filepath.Dir(tmpfilePath), updateStore.name)
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -379,31 +379,51 @@ func repackSdimg(partitions []partition, image string) error {
 	return nil
 }
 
-func getCandidatesForModify(path string) ([]partition, bool, error) {
-	isArtifact := false
+type CandidateType int
+
+const (
+	Unknown CandidateType = iota
+	ModuleImageArtifact
+	RootfsImageArtifact
+	RawSDImage
+)
+
+func getCandidatesForModify(path string) (CandidateType, []partition, error) {
 	modifyCandidates := make([]partition, 0)
 
 	// first we need to check  if we are having artifact or image file
 	art, err := os.Open(path)
 	if err != nil {
-		return nil, isArtifact, errors.Wrap(err, "can not open artifact")
+		return Unknown, nil, errors.Wrap(err, "can not open artifact")
 	}
 	defer art.Close()
 
-	if err = validate(art, nil); err == nil {
-		// we have VALID artifact, so we need to unpack it and store header
-		isArtifact = true
+	var foundType = Unknown
+	aReader := areader.NewReader(art)
+	r, err := read(aReader, nil, nil)
+	if err == nil {
+		// we have VALID artifact,
+
+		// First check if it is a module-image type
+		inst := r.GetHandlers()
+		if len(inst) > 0 && inst[0].GetUpdateType() != "rootfs-image" {
+			return ModuleImageArtifact, nil, nil
+		}
+
+		// Else, it must be a rootfs-image so extract the partitions
+		foundType = RootfsImageArtifact
 		rawImage, err := unpackArtifact(path)
 		if err != nil {
-			return nil, isArtifact, errors.Wrap(err, "can not process artifact")
+			return Unknown, nil, errors.Wrap(err, "can not process artifact")
 		}
 		modifyCandidates = append(modifyCandidates, partition{path: rawImage})
 	} else {
+		foundType = RawSDImage
 		parts, err := processSdimg(path)
 		if err != nil {
-			return nil, isArtifact, errors.Wrap(err, "can not process image file")
+			return Unknown, nil, errors.Wrap(err, "can not process image file")
 		}
 		modifyCandidates = append(modifyCandidates, parts...)
 	}
-	return modifyCandidates, isArtifact, nil
+	return foundType, modifyCandidates, nil
 }
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -39,8 +39,12 @@ func Cat(c *cli.Context) (err error) {
 	}
 	r, err := virtualPartitionFile.Open(comp, c.Args().First())
 	defer func() {
-		if r != nil {
-			r.Close()
+		if r == nil {
+			return
+		}
+		cerr := r.Close()
+		if err == nil {
+			err = cerr
 		}
 	}()
 	if err != nil {
@@ -169,6 +173,9 @@ func Install(c *cli.Context) (err error) {
 
 func Remove(c *cli.Context) (err error) {
 	wclose := func(w io.Closer) {
+		if w == nil {
+			return
+		}
 		cerr := w.Close()
 		if err == nil {
 			err = cerr
--- a/cli/mender-artifact/main.go
+++ b/cli/mender-artifact/main.go
@@ -310,7 +310,7 @@ func getCliContext() *cli.App {
 		},
 		cli.StringFlag{
 			Name: "verification-key, v",
-			Usage: "Full path to the public verification key that is used by the client  " +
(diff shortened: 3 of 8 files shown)


lluiscampos:
Please have a look @oleorhagen (also to the other PR if you have the time)

cli/mender-artifact/main.go
```diff
@@ -56,6 +56,18 @@ func getCliContext() *cli.App {
 	app.Author = "mender.io"
 	app.Email = "contact@mender.io"
 
+	privateKeyFlag := cli.StringFlag{
+		Name: "key, k",
+		Usage: "Full path to the public key that will be used to verify " +
```
olepor:
```suggestion
		Usage: "Full path to the private key that will be used to verify " +
```

cli/mender-artifact/copy.go
```diff
@@ -37,10 +37,14 @@ func Cat(c *cli.Context) (err error) {
 	if !isimg.MatchString(c.Args().First()) {
 		return cli.NewExitError("The input image does not seem to be a valid image", 1)
 	}
-	r, err := virtualPartitionFile.Open(comp, c.Args().First(), c.String("key"))
+	r, err := virtualPartitionFile.Open(comp, c.Args().First())
 	defer func() {
-		if r != nil {
-			r.Close()
+		if r == nil {
+			return
```
olepor: Nice. How it should have been done in the first place.

cli/mender-artifact/artifacts.go
```diff
@@ -159,10 +158,16 @@ func unpackArtifact(name string) (string, error) {
 		return "", errors.Wrap(err, "failed to register install handler")
 	}
 
-	err = aReader.ReadArtifact()
+	r, err := read(aReader, nil, nil)
 	if err != nil {
 		return "", err
 	}
+
+	inst := r.GetHandlers()
```
olepor: Maybe `getUpdateType` would later prove handy as a function call?

cli/mender-artifact/modify.go
```diff
@@ -43,25 +41,22 @@ func modifyArtifact(c *cli.Context) error {
...
 
 	if err != nil {
 		return cli.NewExitError("Error selecting images for modification: "+err.Error(), 1)
 	}
 	// strip the data and boot partitions
-	if isArtifact {
-		modifyCandidates = modifyCandidates[0:1]
-		for _, mc := range modifyCandidates {
-			defer os.Remove(mc.path)
+	if modifyCandidates != nil {
```
olepor: Curious as to why this check is necessary, since the error is already handled above?

cli/mender-artifact/modify.go
```diff
@@ -81,8 +76,12 @@ func modifyArtifact(c *cli.Context) error {
 
 	if isArtifact {
 		// re-create the artifact
-		err := repackArtifact(comp, c.Args().First(), modifyCandidates[0].path,
-			c.String("key"), c.String("name"))
+		rootfs := ""
+		if modifyCandidates != nil {
```
olepor: What happens if there are no partitions to repack? (ie, modifyCandidates is nil? )

cli/mender-artifact/partition.go
```diff
@@ -57,20 +57,24 @@ type vFile int
...
 	imgname, fpath, err := parseImgPath(imgpath)
 	if err != nil {
 		return nil, err
 	}
-	modcands, isArtifact, err := getCandidatesForModify(imgname, []byte(key))
+	modcands, isArtifact, err := getCandidatesForModify(imgname)
 	if err != nil {
 		return nil, err
 	}
+	if modcands == nil {
```
olepor: Same as the others. Why isn't no partitions returned an error?

cli/mender-artifact/modify.go
lluiscampos: I see now that my commit message is not as clear as I hoped for. I will try to clarify it here (this applies to other of your comments in the PR).

I have introduced a design change on this function: now it can return `nil` partitions on valid Artifacts that just happen to have no partitions (like Update Modules). Hence there is no error returned, but the caller still needs to check if there are partitions or not.

This is the detail I wanted to discuss in person.

lluiscampos: I we do this in order to support changing the name (and maybe in the future other attributes like device type for instance) on Artifacts that are not rootfs-image type

cli/mender-artifact/modify.go
lluiscampos: It means that we haven't made any change in the partitions, only in the headers of the Artifacts, so no repacking is needed.

cli/mender-artifact/partition.go
lluiscampos: See previous comments.

cli/mender-artifact/main.go
lluiscampos: Hehehe, good catch :)

I will fix it in PR #194 where the commit belongs.

cli/mender-artifact/modify.go
oleorhagen: I've had another look now. I mean, the PR doesnt interfere with the current code in any degrading way  -- which to be fair is good soup to begin with. However, maybe this would be a nice opportunity to make this more explicit. I feel like `getCandidatedForModify` might make this more explicit, maybe through returning some sort of an enum, instead of a bool, where `isArtifact` is currently at.
Thoughts?

lluiscampos: I thought about it at some point, actually. But now that a golang expert (you) suggests it, I will do it ;)

oleorhagen: haha, I'd be careful going down that road...

lluiscampos:
@oleorhagen please have a look at the last commit, cc269e4

cli/mender-artifact/partition.go
```diff
@@ -62,22 +62,22 @@ func (v vFile) Open(comp artifact.Compressor, imgpath string) (VPFile, error) {
...
 	}
-	if isArtifact {
+	if candidateType == RootfsImageArtifact {
 		return newArtifactExtFile(comp, fpath, modcands[0])
+	} else if candidateType == RawSDImage {
+		return newSDImgFile(fpath, modcands)
 	}
-	return newSDImgFile(fpath, modcands)
-
+	return nil, fmt.Errorf("No partitions found on file %s, only " +
```
oleorhagen: `in` file

cli/mender-artifact/partition.go
```diff
@@ -62,18 +62,22 @@ func (v vFile) Open(comp artifact.Compressor, imgpath string) (VPFile, error) {
...
 		return nil, err
 	}
-	modcands, isArtifact, err := getCandidatesForModify(imgname)
+	candidateType, modcands, err := getCandidatesForModify(imgname)
 	if err != nil {
 		return nil, err
 	}
-	for i := 0; i < len(modcands); i++ {
-		modcands[i].name = imgname
+	if modcands != nil {
```
oleorhagen: VirtualPartitionFiles will never modify artifact metadata like name, so couldn't the test here simply be if it does not contain any partitions. Error?

cli/mender-artifact/artifacts.go
```diff
@@ -379,41 +379,51 @@ func repackSdimg(partitions []partition, image string) error {
...
+
+func getCandidatesForModify(path string) (CandidateType, []partition, error) {
+	var foundType = Unknown
 	modifyCandidates := make([]partition, 0)
 
 	// first we need to check  if we are having artifact or image file
 	art, err := os.Open(path)
 	if err != nil {
-		return nil, isArtifact, errors.Wrap(err, "can not open artifact")
+		return foundType, nil, errors.Wrap(err, "can not open artifact")
```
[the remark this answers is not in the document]
lluiscampos: Yes! Even if it is the same, it will help reading the code

cli/mender-artifact/artifacts.go
lluiscampos: I missed this comment. But I guess it is okey to leave it for the next refactoring day... ;)

cli/mender-artifact/partition.go
lluiscampos: Mmm, right. But then what about the last return?

Because it was bolean, if artifact, else; but now we have the enum, so if artifact, elif sdimg, else?

cli/mender-artifact/partition.go
lluiscampos: :+1:

lluiscampos:
Thanks. What about the naming? I am not very convinced on `candidateType`...

cli/mender-artifact/partition.go
lluiscampos: Ok, I got a suggestion for this. Check the new change set

oleorhagen:
> Thanks. What about the naming? I am not very convinced on `candidateType`...

Yes, I pondered upon it a little bit. I guess I just let it go as the function is named candidate. But maybe something like partitionType? Or maybe payloadType?

cli/mender-artifact/partition.go
oleorhagen: Yes! Way better imo

lluiscampos:
> [...] But maybe something like partitionType? Or maybe payloadType?

It cannot be partitionType because it could be a ModuleImageArtifact, which has no partitions. For payload, an Update Module could have no payload at all.

We'll leave it as it is then, thanks for the thought.

oleorhagen:
Then let's keep it easy :)

lluiscampos:
Tests ok. Rebasing a bit and merging

PR #199, opened 2019-05-21
MEN-2506: Throw a CLI error in case of a copy error. 

oleorhagen:
Changelog: None

### DIFF
--- a/cli/mender-artifact/copy_test.go
+++ b/cli/mender-artifact/copy_test.go
@@ -180,6 +180,14 @@ func TestCopyRootfsImage(t *testing.T) {
 			},
 		},
 		{
+			initfunc: func(imgpath string) {
+				assert.Nil(t, ioutil.WriteFile("input.txt", []byte("dummy-text"), 0755))
+			},
+			name: "copy file into a non-existing directory",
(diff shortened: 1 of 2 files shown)


cli/mender-artifact/copy_test.go
```diff
@@ -179,6 +179,14 @@ func TestCopyRootfsImage(t *testing.T) {
 				assert.Equal(t, "artifact_name=foobar", out.String())
 			},
 		},
+		{
+			initfunc: func(imgpath string) {
+				assert.Nil(t, ioutil.WriteFile("input.txt", []byte("dummy-text"), 0755))
```
lluiscampos: The bug is still present for empty files. The following workflow will reveal it:

```
$ touch hola
$ mender-artifact cp hola artifact.mender:/non/existing/dummy
$ mender-artifact cat artifact.mender:/non/existing/dummy
failed to copy from: artifact.mender:/non/existing/dummy to stdout: err: extFile: ReadError: debugfsCopyFile failed: file /non/existing/dummy not found in image
```

You can also reveal it with a similar Unit Test where the input file is created like:

```
ioutil.WriteFile("input.txt", []byte(""), 0755)
```

PR #200, opened 2019-05-22
Added a build step for macOS to the Travis build.

oleorhagen:
Changelog: Title

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,10 @@ go:
 install:
     # Get tools used in Makefile
     - make get-tools
-    - sudo apt-get install mtools liblzma-dev
+
+addons:
+  apt:
+    packages: mtools liblzma-dev
 
 before_script:
     # Print build info.
@@ -31,12 +34,24 @@ before_script:
     # Get last remaining tags, if any.
     - git fetch --tags origin
 
-    - make extracheck
-
-script:
-    - make coverage
-
-    - make
+jobs:
+  include:
+    - stage: build
+      script:
+          - make build
+          - make extracheck
+      os: linux
+    - stage: build
+      script: make build
+      os: osx
+    - stage: test
+      script: true
+      os: osx
+    - stage: test
+      script:
+          - make coverage
+          - make test
+      os: linux
 
 after_success:
     # Integrate with https://codecov.io
--- a/Dockerfile.binaries
+++ b/Dockerfile.binaries
@@ -8,7 +8,7 @@ RUN mkdir -p /go/src/github.com/mendersoftware/mender-artifact
 WORKDIR /go/src/github.com/mendersoftware/mender-artifact
 RUN go get 'golang.org/x/sys/windows'
 ADD ./ .
-RUN make buildNatives
+RUN make build-natives
 
 FROM alpine:3.9
 RUN apk add xz-dev
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ PLATFORMS := darwin linux windows
 
 GO_LDFLAGS_WIN = -ldflags "-X main.Version=$(VERSION) -linkmode=internal -s -w -extldflags '-static' -extld=x86_64-w64-mingw32-gcc"
 
-buildNatives:
+build-natives:
 	@arch="amd64";
 	@echo "building mac";
 	@env GOOS=darwin GOARCH=$$arch CGO_ENABLED=0 \


.travis.yml
```diff
@@ -33,10 +33,23 @@ before_script:
...
+  include:
+    - stage builds
+    script:
+      make build-natives
+
+    # If it builds for all the required platforms, run the unit tests.
+    - stage tests
+    script:
+    # Unit tests will only be run for linux.
+      - if [ "$TRAVIS_OS_NAME" = "linux" ]; then make coverage && make ; fi
```
kacf: I seem to recall there being some way to make conditionals at the parsing stage, not in the shell command. Have you looked into that? It might save spinning up a whole Mac VM in Travis just to tear it down again immediately.

oleorhagen: Nice! I've been looking for it, and couldn't find it. So I spun this up to see if it did in fact spawn a new VM for the second build step. I was sort of hoping it wouldn't.

.travis.yml
```diff
@@ -33,10 +33,23 @@ before_script:
...
+os:
+  - linux
+  - osx
+  # - windows -- currently not built for windows
+
+matrix:
+  include:
+    - stage builds
+    script:
+      make build-natives
```
mzedel: here you could also skip the osx platform, as the https://github.com/mendersoftware/mender-artifact/blob/master/Dockerfile.binaries should produce this - e.g. using: `docker build -t mender-artifact-binaries -f Dockerfile.binaries .` and then just `docker run --rm --entrypoint "/bin/sh" -v $(pwd):/binary mender-artifact-binaries -c "cp /go/bin/mender-artifact-darwin /binary"`... so you would only use the osx platform to run the tests with the herein built binary

oleorhagen: I thought this was only supposed to check if it builds on osx, and not run the tests.

mzedel: well we could do both on travis - I am not sure what time it costs to spin up osx and if an instance can be reused between build and test, but the dockerfile could build mac (and potentially windows) builds to pass the result straight to a testing stage.

oleorhagen: I think that is contained in https://tracker.mender.io/secure/RapidBoard.jspa?rapidView=10&view=detail&selectedIssue=MEN-2505&quickFilter=67, so I will move to that next.

mzedel: yes, there we would definitely need an osx instance, probably won't hurt the build also runs on travis osx

.travis.yml
```diff
@@ -11,7 +11,12 @@ go:
 install:
     # Get tools used in Makefile
     - make get-tools
-    - sudo apt-get install mtools liblzma-dev
+
+addons:
```
oleorhagen: Moved to using the Travis addons functionality in order to seperate the dependency management. It is this, or doing it in the makefile. I think this is cleaner.

.travis.yml
```diff
@@ -31,12 +36,24 @@ before_script:
     # Get last remaining tags, if any.
     - git fetch --tags origin
 
-    - make extracheck
-
-script:
-    - make coverage
-
-    - make
+jobs:
```
oleorhagen: The new build matrix. 
For now there are two stages, which means that a new VM is booted for each step. Personally, I think we maybe should keep it all in one stage. It will be faster that way. Thoughts?

Dockerfile.binaries
```diff
@@ -8,7 +8,7 @@ RUN mkdir -p /go/src/github.com/mendersoftware/mender-artifact
 WORKDIR /go/src/github.com/mendersoftware/mender-artifact
 RUN go get 'golang.org/x/sys/windows'
 ADD ./ .
-RUN make buildNatives
+RUN make build-natives
```
oleorhagen: Renamed the make dependency to align with the old naming.

.travis.yml
mzedel: IMO we should keep it in 2 stages to avoid having required dependencies present in the test environment that we would otherwise not know about - should be the exception, but still better to catch it in a test than on a users machine

kacf: +1 to what @mzedel said!

.travis.yml
```diff
@@ -31,12 +36,27 @@ before_script:
...
+          - make extracheck
+      os: linux
+    - stage: build
+      script: make build # Builds for all three platforms using the go-cross-compiler (is this necessary?)
+      os: osx
+    - stage: test
+      install: export PATH=/usr/local/opt/e2fsprogs/sbin/debugfs:$PATH # debugfs is not added to the path by default.
+      script:
+          - make coverage
+          - make test # test & extracheck
```
kacf: Ambiguous comment: I read this as `make test` = "Run tests and extra check", but I suppose you mean it more as a TODO? I suggest removing it, since the extracheck only needs to be done on one platform (code analysis), and it's already present in the build stage.

kacf:
> One minor comment comment (comment about a comment!), everything else looks good to me.

Btw, that's also assuming Manuel's comments are addressed.

.travis.yml
oleorhagen: Yes, sry, should have commented, this is WIP. Will be gone by the end!

oleorhagen: Meta-comment :)

PR #201, opened 2019-05-24
MEN-2505: Test the mac-binary on macOS in Travis. 

oleorhagen:
Changelog: None

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,21 +37,22 @@ before_script:
 jobs:
   include:
     - stage: build
+      os: linux
       script:
           - make build
           - make extracheck
-      os: linux
     - stage: build
+      os: osx
       script: make build
-      os: osx
     - stage: test
-      script: true
       os: osx
+      install: make install
+      script: cd ./tests-acceptance-mac/ && ./test-mac.sh
     - stage: test
+      os: linux
       script:
           - make coverage
           - make test
-      os: linux
 
 after_success:
     # Integrate with https://codecov.io
--- a/tests-acceptance-mac/ma-read-output.golden
+++ b/tests-acceptance-mac/ma-read-output.golden
@@ -1,0 +1,22 @@
+Mender artifact:
+  Name: release-1
+  Format: mender
+  Version: 3
+  Signature: no signature
+  Compatible devices: '[beaglebone]'
+  Provides group: 
+  Depends on one of artifact(s): []
+  Depends on one of group(s): []
+  State scripts:
+
+Updates:
+    0:
+    Type:   rootfs-image
+    Provides: Nothing
+    Depends: Nothing
+    Metadata: Nothing
+    Files:
+      name:     rootfs.ext4
+      size:     0
+      modified: 2019-05-24 12:48:19 +0200 CEST
+      checksum: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
--- a/tests-acceptance-mac/ma-write-help-text.golden
+++ b/tests-acceptance-mac/ma-write-help-text.golden
@@ -1,0 +1,13 @@
+NAME:
+   mender-artifact write - Writes artifact file.
+
+USAGE:
+   mender-artifact write command [command options] [arguments...]
+
+COMMANDS:
+     rootfs-image  Writes Mender artifact containing rootfs image
+     module-image  Writes Mender artifact for an update module
+
+OPTIONS:
+   --help, -h  show help
+   
--- a/tests-acceptance-mac/test-mac.sh
+++ b/tests-acceptance-mac/test-mac.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -e
+
+# A simple shell script to verify that an Artifact is capeable of reading and writing
+# an artifact on macOS (see https://tracker.mender.io/browse/MEN-2505).
+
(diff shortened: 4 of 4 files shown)


.travis.yml
```diff
@@ -45,7 +45,13 @@ jobs:
       script: make build
       os: osx
     - stage: test
-      script: true
+      script:
+        - go test -v $(go list ./... | grep -vE "vendor|cli|utils")
```
oleorhagen: More fine-grained control of which tests to run. This passes.
The utils fails cause of the dependency upon /bin/true, which is at a different location on macOS. This could be handled I guess, if you want.
Also, the cli tests are dropped, but run below as a tautology.

.travis.yml
```diff
@@ -45,7 +45,13 @@ jobs:
...
+        # Allow the cli tests to fail, as some of them rely on external tooling,
+        # such as debugfs.
+        # The utils tests are dropped, as it assumes the existence of /bin/true,
+        # which was not found on osx.
+        - go test -v github.com/mendersoftware/mender-artifact/cli/mender-artifact || true
```
oleorhagen: So I guess the question is:
Do you want it handled more precisely? ie, running tests individually,
or rather excluding tests that rely on external binaries manually?

.travis.yml
```diff
@@ -45,7 +45,13 @@ jobs:
       script: make build
       os: osx
     - stage: test
-      script: true
+      script:
+        - go test -v $(go list ./... | grep -vE "vendor|cli|utils")
+        # Allow the cli tests to fail, as some of them rely on external tooling,
+        # such as debugfs.
```
lluiscampos: I don't think this is a good idea.

IMO, the CLI tests are actually the only ones to test for Mac OS, as the CLI tool is exactly what we want to support on this platform. So we shouldn't just ignore all of them and, instead, find a smarter way to skip (via `testing.skip()`) only the problematic tests.

What do you think? Can you convince me of the opposite?

.travis.yml
lluiscampos: Yes! That is my point in the other comment. Sorry I didn't read this before. IMO, a "do whatever || true" cannot be a good idea.

.travis.yml
oleorhagen: I agree, but it will be a time vs effort decision. If we decide to do the fine-grained solution, this will have to be manually tweaked by everyone, every time a new test is added on a mac.
But I guess you are right. 
I'll start cherry-excluding tests.

lluiscampos: Give it a try! Thanks :+1:

.travis.yml
```diff
@@ -37,21 +37,22 @@ before_script:
...
-      os: linux
     - stage: build
-      script: make build
       os: osx
+      script: make build
     - stage: test
-      script: true
       os: osx
+      install: make install
+      script: cd ./travis-scripts/ && ./test-mac.sh
```
oleorhagen: Moved to running only the mender-artifact read,write validate functionality on mac.

travis-scripts/ma-read.golden
```diff
@@ -0,0 +1,22 @@
+Mender artifact:
```
oleorhagen: Added golden files for verifying the output of the mender-artifact commands. These need to be updated given output changes ofc.

travis-scripts/test-mac.sh
```diff
@@ -0,0 +1,23 @@
...
+mender-artifact write | diff ma-write.golden -
+
+########## Step 1 - Write an Artifact  ##########
+mender-artifact write rootfs-image -t beaglebone -n release-1 -f rootfs.ext4 -o artifact.mender
+
+########## Step 2 - Verify an Artifact ##########
+mender-artifact validate artifact.mender > /dev/null
+
+########## Step 3 - Read an Artifact   ##########
+mender-artifact read artifact.mender | diff --ignore-matching-lines='modified:.*' ma-read.golden -
```
oleorhagen: The modified: somedate line is ignored, as obviously the time will differ in this instance.

.travis.yml
lluiscampos: This folder is only for Mac testing, the fact that is in Travis is just an implementation detail (that eventually we will change). I suggest just naming the folder `tests-accpentance-mac` or similar

travis-scripts/ma-write.golden
```diff
@@ -0,0 +1,13 @@
+NAME:
```
lluiscampos: It took me a second to understand the naming of these files. I suggest renaming to something like `ma-read-output.golden` and `ma-write-help-text.golden`, so that it is understandable why are they not symmetric.

travis-scripts/test-mac.sh
lluiscampos: :+1:

oleorhagen:
> Okey, I still think we could have had these in go test framework (and perhaps controlling the tests to compile using golang build constrains), but your work is tidy enough now and we got some basic validation of the binary on Mac OS :)

Yes, I agree. But this will have to be another task, for another time. Let's start out simple for now, then have another look when we have a mac for local testing. I guess the final goal would be incorporate all the unit tests -- also on a mac.

> Just minor comments on the naming of files, follow my suggestions if you wish.

Like the naming! Incorporated into the latest commit.

oleorhagen:
Unrelated integration test failure.

PR #205, opened 2019-06-18
Men 2544

oleorhagen:


oleorhagen:
@lluiscampos This is WIP, but It'd be nice if you had a look. 
Main points are:
* It keeps the file permissions on copy in (copy out we cannot do at the moment, as it does try to change the file owner at the same time, which is not permitted with the permissions that mender-artifact has)
* Refactored the logic for how the test cases are run on the different images (artifact, sdimg, fat-sdimg)
Be great if you had some suggestions pointers for the code at the current stage, even though it is rough :)

cli/mender-artifact/copy_test.go
```diff
@@ -111,18 +109,25 @@ func TestCopyRootfsImage(t *testing.T) {
...
+			argv: []string{"mender-artifact", "cp", "foo.txt", "menderimg.mender:"},
 			err:  "please enter a path into the image",
+			verifyTestFunc: func(ipath string) {
+				os.Remove("foo.txt")
+			},
 		},
 		{
 			name:     "write artifact_info file to stdout",
-			argv:     []string{"mender-artifact", "cat", ":/etc/mender/artifact_info"},
+			argv:     []string{"mender-artifact", "cat", "<artifact|sdimg|fat-sdimg>:/etc/mender/artifact_info"},
```
lluiscampos: I can see what you are doing.

In the one hand it increases the complexity of the test execution below, but in the other hand it improve readability in the command line arguments here in the test cases definitions. Nice work!

cli/mender-artifact/copy_test.go
```diff
@@ -166,59 +172,80 @@ func TestCopyRootfsImage(t *testing.T) {
...
 			},
 		},
 		{
-			name:        "error: artifact does not contain a data partition",
-			validimages: []int{0},
-			argv:        []string{"mender-artifact", "cp", "output.txt", ":/data/test.txt"},
-			err:         "newArtifactExtFile: A mender artifact does not contain a data partition, only a rootfs",
+			name: "error: artifact does not contain a data partition",
+			initfunc: func(string) {
+				assert.Nil(t, ioutil.WriteFile("output.txt", nil, 0755))
```
lluiscampos: This is very good: not to reuse an "output" from a previous test case

I suggest we call it "input.txt" ;)

lluiscampos: Here and in the following test cases

lluiscampos: Or foo.txt or something else. My point is not to be output as that has another meaning in many test cases

cli/mender-artifact/copy_test.go
```diff
@@ -412,14 +435,12 @@ func TestCopyRootfsImage(t *testing.T) {
 		},
 		{
 			name: "Error when deleting a non-empty directory from an image or Artifact",
-			argv: []string{"mender-artifact", "rm", ":/etc/mender/"},
-			err: `debugfsRemoveDir: debugfs: error running command: "rmdir /etc/mender", err: 
-rmdir: directory not empty
-`,
+			argv: []string{"mender-artifact", "rm", "<artifact|sdimg|fat-sdimg>:/etc/mender/"},
+			err:  `debugfsRemoveDir: debugfs: error running command:`,
```
lluiscampos: If now is single line, let's use the regular quotes " "

cli/mender-artifact/partition.go
```diff
@@ -137,15 +139,15 @@ func newSDImgFile(fpath string, modcands []partition) (sdimgFile, error) {
...
 	if strings.HasPrefix(fpath, "/data") {
 		// The data dir is not a directory in the data partition
-		fpath = strings.TrimPrefix(fpath, "/data")
+		fpath = strings.TrimPrefix(fpath, "/data/")
 		delPartition = append(delPartition, modcands[0:3]...)
 		ext, err := newExtFile(fpath, modcands[3]) // Data partition
 		return sdimgFile{ext}, err
 	}
-	reg := regexp.MustCompile("/(uboot|boot/(efi|grub))")
+	reg := regexp.MustCompile("/(uboot|boot/(efi|grub))[/]")
```
lluiscampos: Good catch

oleorhagen:
@lluiscampos Could you have anotherl look at this ?

oleorhagen:
@kacf, since @lluiscampos is gone now, would you mind having a look? I'd like to merge this finally. The codecov patch is annoying, as the absolute coverage is up! The integration tests should not be affected by this either...

lluiscampos:
Hi @oleorhagen 

I did approve it and AFAICS there are no changes since then. Merging.

PR #206, opened 2019-06-20
Qa 75 production gitlab pipeline

mzedel:


.gitlab-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+cache:
+  paths:
+    - /go/src/github.com
+    - /go/src/golang.org
+    - /go/src/google.golang.org
+    - /go/src/gopkg.in
+
+variables:
+   DOCKER_HOST: tcp://docker:2375/
+   DOCKER_DRIVER: overlay2
```
lluiscampos: What is this for? Do I need it in the Mender QA pipeline?

.gitlab-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+    - apk add bash e2fsprogs e2fsprogs-extra gcc libc6-compat mtools musl-dev parted perl-utils xz-dev
+    - make extracheck
+    - make coverage
+    - mv /go/src/github.com/mendersoftware/mender-artifact/coverage.txt $CI_PROJECT_DIR/coverage.txt
+  artifacts:
+    untracked: true
+    paths:
+      - coverage.txt
+
+build:
```
lluiscampos: We can call this job `build:docker`, so that it is consistent with `build:make` below

.gitlab-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+services:
+  - docker:dind
+
+stages:
+  - test
+  - build
+  - publish
+  - integration
+
+before_script:
```
lluiscampos: This `before_scripts` seems to be used only in `test:static`.

I suggest: making it generic so that it can be reused or moving it directly into test:static (and then refactor a bit the other steps so that they have a proper before_script as well)

awriter/signer.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2019 Northern.tech AS
```
lluiscampos: This file(s) should not be part of the PR. Please rebase from mendersoftware/master

.gitlab-ci.yml
```diff
@@ -0,0 +1,124 @@
...
+  stage: publish
+  dependencies:
+    - build
+  script:
+    - echo "publishing image to docker hub"
+    # - docker load -i image.tar
+    # - if [ ! -z "$CI_COMMIT_TAG" ]; then export IMAGE_TAG=$CI_COMMIT_TAG; else export IMAGE_TAG=$CI_COMMIT_REF_NAME; fi ;
+    # - docker tag $DOCKER_REPOSITORY:pr $DOCKER_REPOSITORY:$IMAGE_TAG ;
+    # - docker login --username=$DOCKER_HUB_USERNAME --password=$DOCKER_HUB_PASSWORD ;
+    # - docker push $DOCKER_REPOSITORY:$IMAGE_TAG ;
```
lluiscampos: Very nice!

lluiscampos: But we'll need to see if we do this from here or from the Mender QA pipeline

.gitlab-ci.yml
mzedel: that used to be required by Gitlab for dind to work, it is now the default volume driver so I'll remove

.gitlab-ci.yml
mzedel: yep, we could also only push on tags, as that will be directly connected to the QA steps, but in the QA pipeline we have all the images in one docker login (one repo with login information to care)

awriter/signer.go
mzedel: tried this, unfortunately rebasing doesn't help - I guess I'll just close this one and reopen from a new branch

.gitlab-ci.yml
kacf: I think we should not push any images from here, but from Jenkins, since that is where our releases are made. Better to have things in one place when we release.

mzedel: Agree, we should use the QA work to align where we publish what - some repos publish from travis and some rely on Jenkins, some on tags, some on master, ...
If we ever get rid of Jenkins we could move all the build-artifact-publishing to the QA pipeline 😄

kacf: Yes, perhaps I should say "mender-qa" pipeline, not Jenkins pipeline. The point is that it's the big one that combines the repos.

PR #208, opened 2019-07-15
Enable signing of artifacts larger than 1MiB

oleorhagen:
Changelog: None

### DIFF
--- a/awriter/signer.go
+++ b/awriter/signer.go
@@ -31,7 +31,6 @@ func SignExisting(src io.Reader, dst io.Writer, key []byte, overwrite bool) erro
 	var foundManifest bool
 	rTar := tar.NewReader(src)
 	wTar := tar.NewWriter(dst)
-	buf := make([]byte, 1048576) // 1MiB
 	for {
 		header, err := rTar.Next()
 		if err == io.EOF {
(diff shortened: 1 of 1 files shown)


awriter/signer.go
```diff
@@ -61,18 +60,9 @@ func SignExisting(src io.Reader, dst io.Writer, key []byte, overwrite bool) erro
...
 
-		read, err := rTar.Read(buf)
-		if err != nil && err != io.EOF {
-			return errors.Wrap(err, "Could not read tar file entry")
-		} else if read == 0 {
-			continue
-		}
-
-		written, err := wTar.Write(buf[:read])
+		_, err = io.Copy(wTar, rTar)
```
oleorhagen: Better to use the io.Copy functionality than basically juggle readers and writers ourself imo.

kacf: :+1:

oleorhagen:
@kacf, It fails the integration tests with the issues we've been seeing lately. Also, the codecov patch is a little weird. 
Is it good to go you reckon?

kacf:
Yes, nothing here I haven't already seen elsewhere. Good to go as far as I'm concerned!

PR #215, opened 2019-09-04
MEN-2745: 'mender-artifact cp' now requires '-' in order to read stdin

oleorhagen:
Changelog: Title

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -93,7 +93,7 @@ func Copy(c *cli.Context) (err error) {
 		return nil
 	case copyinstdin:
 		r = os.Stdin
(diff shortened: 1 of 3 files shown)


oleorhagen:
Now, mender-artifact requires the '-' on the command line in order to read from stdin. There is no check whether or not data is pending on stdin, so empty stdin will now lock. But that is not too surprising I think?

kacf:
No, that is how I would expect it to work. And it allows the rare, but occasionally useful use case of typing something directly into the pipe.

oleorhagen:
@kacf , the tests botched codecov for some reason.... Force, or rerun?

kacf:
Nah, tiny change, coverage would not make a difference here anyway.

kacf:
Thanks for the quick handling!

PR #216, opened 2019-09-06
FIX: Buffer writes from stdin, and flush on close

oleorhagen:
@lluiscampos This should solve your issues. Feel free to have a look. I have not added tests for it yet, but they will be here by the end of the day, in the meantime I don't see any issues trying merging this really, if it is still urgent for you

### DIFF
--- a/cli/mender-artifact/partition.go
+++ b/cli/mender-artifact/partition.go
@@ -277,12 +277,19 @@ func (a *artifactExtFile) Close() (err error) {
 	if a == nil {
 		return nil
 	}
-	if a.repack {
-		err = repackArtifact(a.comp, a.name, a.path, "")
-	}
 	if a.tmpf != nil {
+		if a.flush {
+			a.tmpf.Sync()
+			err = debugfsReplaceFile(a.imagefilepath, a.tmpf.Name(), a.path)
+			if err != nil {
+				return err
+			}
+		}
 		a.tmpf.Close()
 		os.Remove(a.tmpf.Name())
+	}
+	if a.repack {
+		err = repackArtifact(a.comp, a.name, a.path, "")
 	}
 	os.Remove(a.path)
 	return err
@@ -293,6 +300,7 @@ type extFile struct {
 	partition
 	imagefilepath string
 	repack        bool     // True if a write has been done
+	flush         bool     // True if Close() needs to copy the file to the image
 	tmpf          *os.File // Used as a buffer for multiple write operations
 }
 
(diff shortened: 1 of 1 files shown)


cli/mender-artifact/partition.go
```diff
@@ -277,13 +277,20 @@ func (a *artifactExtFile) Close() (err error) {
 	if a == nil {
 		return nil
 	}
-	if a.repack {
-		err = repackArtifact(a.comp, a.name, a.path, "")
-	}
 	if a.tmpf != nil {
+		if a.flush {
+			a.tmpf.Sync()
```
kacf: Flush and Sync are two different things: Flush outputs data that the application has buffered in memory, Sync forces the kernel cache to be dumped to disk.

`os.File` doesn't need flushing because everything is a direct write (as opposed to `bufio.Writer`, which does have buffering/flushing). And there is definitely no need for Sync here, as it only decreases performance and provides no benefit (the kernel cache will survive even if your program crashes). Sync is only for system level changes where an application needs to recover from spontaneous reboots, and you aren't going to do that here.

cli/mender-artifact/partition.go
```diff
@@ -396,15 +396,22 @@ func (ef *extFile) Close() (err error) {
 	if ef == nil {
 		return nil
 	}
-	if ef.repack {
-		part := []partition{ef.partition}
-		err = repackSdimg(part, ef.name)
-	}
 	if ef.tmpf != nil {
+		if ef.flush {
+			ef.tmpf.Sync()
```
kacf: Same here, remove `Sync()`.

oleorhagen:
@kacf Thank you for the review! Insightful as usual. I will incorporate your requests directly. 

However, I do not like that the PR is difficult to understand. Basically the only change is that:

* Now, all writes to the virtual partition file is buffered in a temp-file, prior to handing the file over to debugfs. This means that now. Multiple writes will append to the existing temp-file, and upon close, the temp-file will be copied by debugfs to the file-system at hand.

Does that make sense? Otherwise we can do a face to face tomorrow :)

kacf:
> I do not like that the PR is difficult to understand.

Neither do I, but it could just be me being totally blind to something obvious. Let's do the face time as soon as I get in today.

cli/mender-artifact/partition.go
```diff
@@ -497,14 +493,22 @@ func (f *fatFile) Close() (err error) {
 	if f == nil {
 		return nil
 	}
-	if f.repack {
-		p := []partition{f.partition}
-		err = repackSdimg(p, f.name)
-	}
 	if f.tmpf != nil {
+		if f.flush {
+			cmd := exec.Command("mcopy", "-n", "-i", f.path, f.tmpf.Name(), "::"+f.imageFilePath)
```
kacf: Good catch with FAT.

PR #217, opened 2019-09-13
Fix: mender-artifact modify did not clean up the temp-files created

oleorhagen:
Changelog: None

### DIFF
--- a/cli/mender-artifact/modify.go
+++ b/cli/mender-artifact/modify.go
@@ -55,6 +55,11 @@ func modifyArtifact(c *cli.Context) error {
 		}
 	} else if candidateType == RawSDImage {
 		if len(modifyCandidates) == 4 { // sdimg
+			defer func(partitions []partition) {
+				for _, part := range partitions {
+					os.Remove(part.path)
+				}
+			}(modifyCandidates)
 			modifyCandidates = modifyCandidates[1:3]
 		}
 	}
--- a/cli/mender-artifact/modify_existing_test.go
+++ b/cli/mender-artifact/modify_existing_test.go
@@ -104,6 +104,12 @@ func verifySDImg(image, file, expected string) bool {
 	candidateType, modifyCandidates, err :=
 		getCandidatesForModify(image)
 
+	defer func(partitions []partition) {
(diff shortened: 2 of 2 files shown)


cli/mender-artifact/modify.go
```diff
@@ -89,6 +94,14 @@ func modifyArtifact(c *cli.Context) error {
 	return nil
 }
 
+func cleanSDImg(partitions []partition, candidateType CandidateType) {
+	if candidateType == RawSDImage && len(partitions) == 4 {
```
lluiscampos: Instead of splitting in two places the knowledge of "len 4 in partitions means this or that", I would put the cleanup part at line 63, like:
```
} else if candidateType == RawSDImage {
    if len(modifyCandidates) == 4 { // sdimg
        modifyCandidates = modifyCandidates[1:3]
        // HERE: defer cleanup
    }
}
```

It would then be more consistent to do an specific cleanup according to which candidateType you got from getCandidatesForModify

cli/mender-artifact/artifacts.go
```diff
@@ -164,7 +164,7 @@ func unpackArtifact(name string) (string, error) {
 	}
 
 	inst := r.GetHandlers()
-	if ! (len(inst) == 1 && inst[0].GetUpdateType() == "rootfs-image" ) {
+	if !(len(inst) == 1 && inst[0].GetUpdateType() == "rootfs-image") {
```
lluiscampos: I personally think it is better to split style changes from actual code changes. This will ease the git archeology in the future.

lluiscampos: (this is for future reference, you don't need to do it now)

oleorhagen:
I think I'm gonna have a look at rewriting the resource handling related to the tmp-file handling. Thus it will be a part of that future PR. Thought it nicer to keep it seperate. Initially we did not concern ourself with leftover tmp-files, so there is not really any ownership of the tmp-files, it is just all hot-patches, as we started to care about the leaks. I will have a quick look at taking this by the root instead, if it is not to much work.

PR #218, opened 2019-09-16
Update formatting of files that were out of sync with gofmt

oleorhagen:
Changelog: None

### DIFF
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -164,7 +164,7 @@ func unpackArtifact(name string) (string, error) {
 	}
 
 	inst := r.GetHandlers()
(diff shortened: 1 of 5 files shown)


lluiscampos:
If this is important to you, I suggest creating a style check rule in the pipeline (GitLab!). See the `mender-binary-delta` [pipeline](https://github.com/mendersoftware/mender-binary-delta/blob/master/.gitlab-ci.yml#L98) for inspiration.

lluiscampos:
> @lluiscampos: We already [do that](https://github.com/mendersoftware/mender-artifact/blob/master/Makefile#L97-L102), so I'm curious why it didn't catch it. Are the new Golang versions more strict?

Yeah, I saw it yesterday. We discussed it but couldn't get a good conclusion on it.

PR #219, opened 2019-09-18
Do not close the pipe in TestCopyFromStdin twice

lluiscampos:
Port bugfix to 3.1.x

This should have been ported between beta and final 3.1.0... But anyway, let's get it in now so that can be part of 3.1.1


lluiscampos:
@oleorhagen ^

oleorhagen:
I don't understand why it needs to be in the release?

oleorhagen:
Are the builds failing?

lluiscampos:
The release build "pass", because they don't run unit tests. And I found this out now as the tests are failing for the official Debian packages build.

No problem, I have added it as a patch and we will drop the patch in 3.0.1

PR #221, opened 2019-09-23
MEN-2800: fix erroneously report of "-dirty" in the version string.

lluiscampos:
```
Remove .dockerignore so that the full repo is copied to the image

As the Dockerfile itself was not being sent to the Docker image, git in
the running container will detect it as a deleted file, triggering then
a "-dirty" suffix in the version string.

Changelog: MEN-2800: fix erroneously report of "-dirty" in the version
string.

```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,0 @@
-Dockerfile


lluiscampos:
Reminder: this needs to be ported to 3.1.x release branch

PR #224, opened 2019-09-26
MEN-2156: Remove support for Mender-Artifact format version 1

oleorhagen:
Changelog: Mender-Artifact format version 1 is hereby deprecated,
and neither reading or writing the version 1 of the format is no longer
supported. Please move to using a newer version.

### DIFF
--- a/cli/mender-artifact/validate_test.go
+++ b/cli/mender-artifact/validate_test.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -66,7 +66,7 @@ var validateTests = []struct {
 	validateKey   []byte
 	expectedError error
 }{
-	{1, nil, nil, nil},
+	{2, nil, nil, nil},
 	{2, []byte(PrivateValidateRSAKey), []byte(PublicValidateRSAKey), nil},
 	{2, []byte(PrivateValidateRSAKey), []byte(PublicValidateRSAKeyError),
 		ErrInvalidSignature},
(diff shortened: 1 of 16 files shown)


cli/mender-artifact/write.go
```diff
@@ -70,7 +70,7 @@ func writeRootfs(c *cli.Context) error {
 	var h handlers.Composer
 	switch version {
 	case 1:
-		h = handlers.NewRootfsV1(c.String("file"))
+		return cli.NewExitError("Error: Mender-Artifact version 1 is deprecated", errArtifactUnsupportedVersion)
```
kacf: I would say "not supported". "Deprecated" indicates that it's still possible to use it.

cli/mender-artifact/write.go
```diff
@@ -328,6 +325,10 @@ func writeModuleImage(ctx *cli.Context) error {
 	}
 	version := ctx.Int("version")
 
+	if version == 1 {
+		return cli.NewExitError("Mender-Artifact version 1 is deprecated", 1)
```
kacf: Same here.

cli/mender-artifact/validate_test.go
```diff
@@ -66,7 +66,6 @@ var validateTests = []struct {
 	validateKey   []byte
 	expectedError error
 }{
-	{1, nil, nil, nil},
```
lluiscampos: This is still a valid test case. Upgrade it to v2 instead of removing it

lluiscampos: This might apply to other places as well. I have not reviewed your changes, but while working on [other issue](https://tracker.mender.io/browse/MEN-2802) it occurred to me that you might be just removing this test case.

oleorhagen: How is it still a valid test? It has no key. 
I figured this was its V2 equivalent:
`	{2, []byte(PrivateValidateRSAKey), nil, nil}`

lluiscampos: That is exactly the test case I am fixing. See https://github.com/mendersoftware/mender-artifact/pull/227

oleorhagen:
Cancelled the remaining test-pipeline, as they have passed previously, and no code change (except testing) has been incorporated since the last run.

PR #225, opened 2019-09-27
Update the type-info documentation in the version 3 artifact format

oleorhagen:
This commit updates the description of the values allowed in the type-info
headers in the version 3 of the artifact format. Formerly only the key
`rootfs-image-checksum` was allowed, while now, any key is allowed, with
the only allowed value types being string, or array of strings.

Changelog: Commit

### DIFF
--- a/Documentation/artifact-format-v3.md
+++ b/Documentation/artifact-format-v3.md
@@ -270,19 +270,16 @@ As an opposite to the list of global `artifact_provides` being a part of
 `header-info` file, the `artifact_provides` section in the `type-info` file
 is a set of parameters specific for a given payload type.
 
-The list of currently supported parameters is as follows:
-
-* `rootfs_image_checksum` is the checksum of the image contained within the
-  Artifact
+The `artifact_provides` is a key-value store, where the value is either a string,
+or an array of strings.
 
 #### artifact_depends
 
 The `artifact_depends` section in the `type-info` file is a set of parameters
-specific for a given payload type. The list of currently supported
-parameters is as follows:
+specific for a given payload type.
 
-* `rootfs_image_checksum` is the checksum of the image that needs to be installed
-on the device before current Artifact can be installed
+The `artifact_depends` is a key-value store, where the value is either a string,
+or an array of strings.
 
 
 ### meta-data
@@ -381,11 +378,6 @@ These files and attributes are allowed:
     },
   }
   ```
-
-At the moment ONLY a `type-info` file is allowed which can contain only
-`artifact_depends` with one field: `rootfs_image_checksum` parameters and type
-of the payload.
-
 
 data
 ----


oleorhagen:
@kacf, please have a look, and see if this is like you described

Documentation/artifact-format-v3.md
```diff
@@ -270,19 +270,17 @@ As an opposite to the list of global `artifact_provides` being a part of
...
 
 #### artifact_depends
 
 The `artifact_depends` section in the `type-info` file is a set of parameters
 specific for a given payload type. The list of currently supported
 parameters is as follows:
 
-* `rootfs_image_checksum` is the checksum of the image that needs to be installed
-on the device before current Artifact can be installed
+The `aritfact_depends` is a key-value store, where the value is either a list,
```
kacf: `s/aritfact_depends/artifact_depends/`

Documentation/artifact-format-v3.md
```diff
@@ -270,19 +270,17 @@ As an opposite to the list of global `artifact_provides` being a part of
...
-* `rootfs_image_checksum` is the checksum of the image contained within the
-  Artifact
+The `artifact_provides` is a key-value store, where the value is either a list,
+or an array of strings.
 
 #### artifact_depends
 
 The `artifact_depends` section in the `type-info` file is a set of parameters
 specific for a given payload type. The list of currently supported
 parameters is as follows:
```
kacf: Maybe lose the last sentence here.

Documentation/artifact-format-v3.md
```diff
@@ -383,9 +381,8 @@ These files and attributes are allowed:
   ```
 
 At the moment ONLY a `type-info` file is allowed which can contain only
-`artifact_depends` with one field: `rootfs_image_checksum` parameters and type
-of the payload.
-
+`artifact_depends` where each key is a string, and each value is either a
+string, or a list of strings.
```
kacf: Actually, as planned it can take on any value that the original `type-info` file can have (this was written quite a long time ago). However, augmented artifacts are rejected by the client currently, so it's not actually in use. Until we start using it, I would remove this whole paragraph. No need to mention any restrictions at this point.

Documentation/artifact-format-v3.md
oleorhagen: You don't want the value type constraints mentioned at all?

kacf: No, I do. I mean literally just this sentence: "The list of currently supported parameters is as follows:"

Documentation/artifact-format-v3.md
```diff
@@ -270,19 +270,16 @@ As an opposite to the list of global `artifact_provides` being a part of
 `header-info` file, the `artifact_provides` section in the `type-info` file
 is a set of parameters specific for a given payload type.
 
-The list of currently supported parameters is as follows:
-
-* `rootfs_image_checksum` is the checksum of the image contained within the
-  Artifact
+The `artifact_provides` is a key-value store, where the value is either a list,
+or an array of strings.
```
kacf: Lol, either I missed this the first time, or it was rewritten. "..either a list, or an array...". Aren't those two the same thing? I think the first one should be string.

Documentation/artifact-format-v3.md
```diff
@@ -270,19 +270,16 @@ As an opposite to the list of global `artifact_provides` being a part of
...
 
 The `artifact_depends` section in the `type-info` file is a set of parameters
-specific for a given payload type. The list of currently supported
-parameters is as follows:
+specific for a given payload type.
 
-* `rootfs_image_checksum` is the checksum of the image that needs to be installed
-on the device before current Artifact can be installed
+The `artifact_depends` is a key-value store, where the value is either a list,
+or an array of strings.
```
kacf: Same here.

Documentation/artifact-format-v3.md
oleorhagen: Lol, probably was.

PR #227, opened 2019-09-30
MEN-2802: Do not validate signed Artiafcts without a key

lluiscampos:
Changelog: mender-artifact will now fail to validate a signed artifact
if no validation key is specified. No behaviour change for unsigned
artifacts.

### DIFF
--- a/cli/mender-artifact/validate.go
+++ b/cli/mender-artifact/validate.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Northern.tech AS
+// Copyright 2019 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cli/mender-artifact/validate.go
```diff
@@ -59,6 +59,10 @@ func validate(art io.Reader, key []byte) error {
 		Log.Debug("key was specified but no digital signature was found")
 		return ErrInvalidSignature
 	}
+	if !keyIsSpecified && ar.IsSigned {
+		Log.Debug("no key was specified for a signed Artifact validation")
+		return ErrInvalidSignature
```
kacf: I think all three of these conditions should have separate errors. And IMHO they should be more descriptive than currently, like "Missing key", "Missing signature" and "Signature is invalid", or something like that.

lluiscampos: I completely agree :smiley: 

My only excuse is that it was already simplified here and in the tests, and I decided on not modifying the existing design... However, I will re-work it now with your encouragement.

lluiscampos: I reworked it a bit. Let me know what you think now.

oleorhagen:
The artifact version 1 removal is now merged, so this can be rebased and merged also

lluiscampos:
Thanks. Done.

PR #229, opened 2019-09-30
Enable typeinfo artifact-depends/provides string and []string values

oleorhagen:
Previously the artifact-depends key in the type-info header was restricted
to contain a single key `rootfs-image-checksum`. This restriction has now
been lifted, and the key can now contain arbitrary string, and []string values.

Changelog: Commit

### DIFF
--- a/artifact/metadata.go
+++ b/artifact/metadata.go
@@ -17,6 +17,7 @@ package artifact
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


oleorhagen:
@kacf, please note that this is WIP, I just wanted ur eyes on it.

oleorhagen:
I decided to stick with the `typeinfo` type, and add a unmarshaler interface to it. The unmarshaling is then verified to be correct during parsing.

Also, I feel the code in itself is clonky, and includes a lot of back and forth with interfaces and type switches. Any good ideas are welcome here

artifact/metadata.go
```diff
@@ -287,9 +288,105 @@ func (ti *TypeInfo) Write(p []byte) (n int, err error) {
 	return len(p), nil
 }
 
-type TypeInfoDepends map[string]string
+type TypeInfoDepends map[string]interface{}
 
-type TypeInfoProvides map[string]string
+func (t TypeInfoDepends) Map() map[string]interface{} {
+	return map[string]interface{}(t)
```
kacf: What's the motivation behind this function? You're not using it anywhere, and this appears to return a copy. Isn't it better to let the user cast and decide whether to copy or not?

Same with the other `Map()` function.

oleorhagen: It is used in the code for `MEN-2549`. Perhaps I should add it in there instead ;)

kacf: If you know where you will use it, then I'm fine with it.

PR #233, opened 2019-10-02
QA-81: GitLab Pipeline completition

lluiscampos:
Introduces the final touches for GitLab migration:

- Publishing on S3 for master binaries (the release responsibility will still be on mender-qa)
- Publishing of coverage reports
- Removal of Travis file

After review, Travis integration will be removed from GitHub configuration (so that the PRs do not "expect" a Travis CI report.


lluiscampos:
@mzedel @kacf Please have a look.

This will be the setup for components with binaries deliverables (`mender-artifact` and `mender-cli`), while for the ones with Docker deliverables we will similarly publish them from master only.

.gitlab-ci.yml
```diff
@@ -57,62 +59,43 @@ build:docker:
...
-    # - docker push $DOCKER_REPOSITORY:$COMMIT_TAG;
+    - for bin in mender-artifact-darwin mender-artifact-linux mender-artifact-windows.exe; do
+        platform=${bin#mender-artifact-};
+        platform=${platform%.*};
+        echo "Publishing master version for ${platform} to S3";
+        aws s3 cp ${bin}
+          s3://$S3_BUCKET_NAME/$S3_BUCKET_PATH/master/${platform}/mender-artifact;
+      done
   only:
     - master
```
kacf: Suggest using:
```
only:
  - branches
  - /^(master|[0-9]+\.[0-9]+\.x)$/
```

Which matches all of our release branches and master.

.gitlab-ci.yml
```diff
@@ -57,62 +59,43 @@ build:docker:
...
-    # - docker load -i image.tar
-    # - export COMMIT_TAG="$CI_COMMIT_REF_NAME"_"$CI_COMMIT_SHA";
-    # - docker tag $DOCKER_REPOSITORY:pr $DOCKER_REPOSITORY:$COMMIT_TAG;
-    # - docker push $DOCKER_REPOSITORY:$COMMIT_TAG;
+    - for bin in mender-artifact-darwin mender-artifact-linux mender-artifact-windows.exe; do
+        platform=${bin#mender-artifact-};
+        platform=${platform%.*};
+        echo "Publishing master version for ${platform} to S3";
+        aws s3 cp ${bin}
+          s3://$S3_BUCKET_NAME/$S3_BUCKET_PATH/master/${platform}/mender-artifact;
```
kacf: If you do the below, then you also need to change both of these `master` references to `${CI_COMMIT_REF_NAME}`.

.gitlab-ci.yml
lluiscampos: So your suggestion is to publish both master or release branch (like 1.2.x) but not exact tags, right?

I like the suggestion. It will save us trouble :)

.gitlab-ci.yml
```diff
@@ -57,62 +59,43 @@ build:docker:
     - docker
 
 build:make:
-  image: golang:1.11
+  image: docker
+  before_script:
+    - apk update && apk add make
```
mzedel: `- apk add --no-cache make` might be faster on alpine based containers, as it doesn't require the update step - only as a FYI, since you plan to use this as a reference

lluiscampos: This is the reference to replicate. So yes, thanks!

.gitlab-ci.yml
mzedel: or would maybe
```
- master
- tags
```
also fit here? otherwise we might have a release with some different name at some point

lluiscampos: The publishing of artifacts from tags should still be done in the mender-qa script as part of the release procedure

lluiscampos:
@kacf It should be okey to be merged now.

PR #238, opened 2019-10-22
CLI help printer adjusts to terminal width

alfrunes:
Overwrite the default help printer with one that aligns the last column
of the ordinary help printer.

changelog: none

### DIFF
--- a/cli/mender-artifact/main.go
+++ b/cli/mender-artifact/main.go
@@ -15,13 +15,17 @@
 package main
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"strings"
 
 	"github.com/mendersoftware/mender-artifact/artifact"
 
+	log "github.com/sirupsen/logrus"
 	"github.com/urfave/cli"
+	"golang.org/x/crypto/ssh/terminal"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cli/mender-artifact/main.go
```diff
@@ -448,9 +452,58 @@ func getCliContext() *cli.App {
...
 }
+
+func upgradeHelpPrinter(defaultPrinter func(w io.Writer, templ string, data interface{})) func(
+	w io.Writer, templ string, data interface{}) {
+	// Applies the ordinary help printer with column post processing
+	return func(stdout io.Writer, templ string, data interface{}) {
+		var buf bytes.Buffer
+		defaultPrinter(&buf, templ, data)
+		const alpha = "abcdefghijklmnopqrstuvxyz"
+		terminalWidth, _, err := terminal.GetSize(int(os.Stdin.Fd()))
```
kacf: `Stdout` should be checked, not `Stdin`.

cli/mender-artifact/main.go
```diff
@@ -448,9 +452,58 @@ func getCliContext() *cli.App {
...
+			newLine := line
+			indent := strings.LastIndex(
+				line[:terminalWidth], "  ")
+			indent += strings.IndexAny(strings.ToLower(line[indent:]),
+				alpha) - 1
+			// find indentation of last column
+			if indent >= terminalWidth-1 ||
+				indent == -1 {
+				indent = 0
+			}
```
kacf: I think the `line[indent:]` will panic if the first expression results in `-1`. This can happen for very narrow terminals. So that needs to be checked immediately after the first expression.

Also I think the second expression will not work correctly if the first characters are not alphanumeric. I don't fully understand why it is needed at all, but would it be better to `TrimLeft` with spaces as the set?

cli/mender-artifact/main.go
```diff
@@ -448,9 +452,58 @@ func getCliContext() *cli.App {
...
+			// Format the last column to be aligned
+			for len(newLine) > terminalWidth {
+				// find word to insert newline
+				idx := strings.LastIndex(newLine[:terminalWidth], " ")
+				if idx == indent {
+					idx = terminalWidth
+				}
+				stdout.Write([]byte(newLine[:idx] + "\n"))
```
kacf: Same here, `idx` can be `-1` for a long last word / narrow terminal.

cli/mender-artifact/main.go
alfrunes: This was just intended for a special case for the mender help menu.
The addition to the indent aligns text that are prefixed with non-alphanumeric characters such as enumerations, e.g.
```
Lorem ipsum etc.
     -- this is a list entry and should continue at
        this indentation level, not
     here.
```

kacf: Ah, ok, smart!

My other point is still valid though.

alfrunes: oops.. I only made the changes for the client PR, and forgot to update this one. I did some minor changes to improve readability as well.

oleorhagen:
ping

alfrunes:
pong

alfrunes:
@lluis can we ignore the codecov patch requirement for this PR?
The `full_integration` fails for completely unrelated reasons.

lluiscampos:
> @lluis can we ignore the codecov patch requirement for this PR?
> The `full_integration` fails for completely unrelated reasons.

Ok

PR #239, opened 2019-10-30
MEN-2880: Make artifact install respect the given file permissions

oleorhagen:
The bug was related to wrong file permissions on the files which were copied over to the target using debugfs. In fact, these were always 0600, as the temp-files did not even have the permissions changed. 

This PR changes that fact, although this implementation did also stumble into some issues (umask overriding the permissions set on creation of the temp-file.

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -17,7 +17,9 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"regexp"
 
 	"github.com/mendersoftware/mender-artifact/artifact"
@@ -148,8 +150,6 @@ func Install(c *cli.Context) (err error) {
 		return cli.NewExitError("Unable to load key: "+err.Error(), 1)
 	}
 
-	var r io.ReadCloser
-	var w io.WriteCloser
 	wclose := func(w io.Closer) {
 		if w == nil {
 			return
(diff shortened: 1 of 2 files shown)


cli/mender-artifact/copy.go
```diff
@@ -261,3 +275,44 @@ func parseCLIOptions(c *cli.Context) int {
...
+	Log.Debugf("Tempfile name: %s\n", name)
+
+	s, _ := tf.Stat()
+	err = tf.Close()
+	if err != nil {
+		return "", err
+	}
+
+	Log.Debugf("The tempfile: %s with permissions: %v\norg-perm: %s\nmode: %s\n",
+		name, s, perm, s.Mode())
```
lluiscampos: What is `org-perm` here? Isn't it the desired `FileMode`? I think it is confusing to mix permissions and mode

cli/mender-artifact/copy.go
```diff
@@ -261,3 +275,44 @@ func parseCLIOptions(c *cli.Context) int {
 		return parseError
 	}
 }
+
+// createTmpFileWithPerm Takes a file, and creates a temp-file copy of the
+// current file, with the permissions given by perm.
+func createTmpFileWithPerm(f *os.File, perm os.FileMode) (string, error) {
```
lluiscampos: An slightly cleaner interface can be `createTmpFileWithPerm` taking just the filename (aka `f.Name()` and the permissions, then returning a filename (as it is doing) instead of having a full file object as the input, an string as the output.

cli/mender-artifact/copy_test.go
```diff
@@ -300,14 +300,96 @@ func TestCopyRootfsImage(t *testing.T) {
...
+					out, err := cmd.CombinedOutput()
+					require.Nil(t, err)
+					assert.True(t, strings.Contains(string(out), "Mode:  0777"))
+				default:
+					t.Fatal("Unexpected file type")
+				}
+			},
+		},
+		{
+			name: "Install file, standard permissions (0700)",
```
lluiscampos: In both tests we say "standard permissions". However, they differ (0777 and 0700)

cli/mender-artifact/copy_test.go
```diff
@@ -300,14 +300,96 @@ func TestCopyRootfsImage(t *testing.T) {
...
+			initfunc: func(imgpath string) {
+				require.Nil(t, ioutil.WriteFile("testkey", []byte("foobar"), 0644))
+			},
+			argv: []string{"mender-artifact", "install", "-m", "0700", "testkey", "<artifact|sdimg|fat-sdimg>:/etc/mender/testkey.key"},
+			verifyTestFunc: func(imgpath string) {
```
lluiscampos: It seems to me that the verifyTestFunc is basically the same as above except for the expected permission bits... Room for refactoring?

Minor suggestion: I don't really mind code duplication in tests.

lluiscampos:
The build error is very strange and it doesn't seem related to the code changes at all...

cli/mender-artifact/copy.go
oleorhagen: Hmm, yeh. True. Rewording

cli/mender-artifact/copy.go
oleorhagen: Agreed!

cli/mender-artifact/copy_test.go
oleorhagen: Fixing.

cli/mender-artifact/copy_test.go
oleorhagen: Yeh, I think I'll leave this one as is actually.

cli/mender-artifact/copy.go
oleorhagen: On second thoughts, I will leave this, as the file contents are also copied over here, thus passing a string will mean, closing the file. Passing the string, then re-opening it in the function.

lluiscampos: Oh, you are right sir!

sbaxter-izo:
Thanks for addressing this so quickly! I'm loving how responsive you all have been :)

oleorhagen:
> Thanks for addressing this so quickly! I'm loving how responsive you all have been :)

Well, it's not merged yet :p 

Backporting will also be a bit more work. But if you are fine with using master, then this should be available to you pretty soon @sbaxter-izo

sbaxter-izo:
> Backporting will also be a bit more work. But if you are fine with using master, then this should be available to you pretty soon @sbaxter-izo

It's not blocking us at the moment as we wrote a quick patch that gets us the permissions of the source file being copied (instead of always installing `0600`), but good to know! We'll definitely drop that in favor of this once it's backported.

PR #241, opened 2019-10-31
[Cherry] MEN-2880: Make artifact install respect the given file permissions 

oleorhagen:
The current Mender-Artifact install functionality had a bug in which the
temp-files it creates and uses for moving a file from the host to the target
with debugfs were created with the standard file permissions (0600). This did
pass test as only (0600) was tested. This fix creates a new temp-file, and
explicitly sets the file-permissions on it to match the one given on the command
line.

It has also added a couple of more test cases (0777, 0700) to make sure that the
same error as seen before does not slip unnoticed through the tests.

Changelog: Title

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -17,7 +17,9 @@ package main
 import (
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"regexp"
 
 	"github.com/mendersoftware/mender-artifact/artifact"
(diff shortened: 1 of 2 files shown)


oleorhagen:
Acceptance tests fail, due to a missing version guard in the acceptance tests for the mender-artifact version. Fix: https://github.com/mendersoftware/meta-mender/pull/854

oleorhagen:
Green build: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/93735892

PR #243, opened 2019-11-08
FIX: MergeArtifact[Depends|Provides] panic on empty depends/provides …

alfrunes:
…in type-info

changelog: none

### DIFF
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -1041,6 +1041,10 @@ func (ar *Reader) MergeArtifactDepends() (map[string]interface{}, error) {
 	retMap := make(map[string]interface{})
 
 	depends := ar.GetArtifactDepends()
+	if depends == nil {
+		// Artifact version < 3
+		return nil, nil
+	}
 
 	retMap["artifact_name"] = depends.ArtifactName
 	retMap["compatible_devices"] = depends.CompatibleDevices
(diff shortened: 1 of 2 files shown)


areader/reader.go
```diff
@@ -1052,6 +1052,8 @@ func (ar *Reader) MergeArtifactDepends() (map[string]interface{}, error) {
 		deps, err := upd.GetUpdateDepends()
 		if err != nil {
 			return nil, err
+		} else if deps == nil {
```
oleorhagen: Question: I think in general that a nil error and type should not be returned from a `Get`, would it not be better to handle this lower in the call-stack?

alfrunes: This isn't an error, the type-info Provides/Depends fields aren't required by the v3 specification.

kacf: Good point. I think originally it was like this to represent the lack of this information from artifact format v2, but there's no reason it can't be empty map.

kacf: I guess it's more about convenience than correct-ness. I can see that it's more convenient if it returns a map always, since in this context an empty map carries the same meaning as a nil pointer, but I don't really have a strong opinion which should be preferred.

oleorhagen: > I guess it's more about convenience than correct-ness. I can see that it's more convenient if it returns a map always, since in this context an empty map carries the same meaning as a nil pointer, but I don't really have a strong opinion which should be preferred.

Yup, my thoughts exactly. But no strong opinions here.

alfrunes:
@kacf, @oleorhagen; I also added a check for calling `Merge...` with version < 3, and return `nil, nil` in such a case.
BTW: the tests failed last iteration because of a corrupted `sstate-cache`, I'm currently investigating this.

alfrunes:
I added unit test for these specific cases as well.

PR #245, opened 2019-12-05
MEN-2956: Write `rootfs_image_checksum` provides in rootfs-image

oleorhagen:
Previously the `write rootfs-image` command did not write the provided checksum
to the Artifact.

This commits makes sure the checksum is always provided when writing a
rootfs-image Artifact.

Changelog: None

### DIFF
--- a/cli/mender-artifact/main.go
+++ b/cli/mender-artifact/main.go
@@ -162,6 +162,12 @@ func getCliContext() *cli.App {
 			Usage: "Full path to the state script(s). You can specify multiple " +
 				"scripts providing this parameter multiple times.",
 		},
+		cli.BoolFlag{
+			Name: "no-checksum-provide",
+			Usage: "Disable writing the provides checksum to the Artifact provides " +
+				"parameters. This is needed in case the targeted devices do not support " +
+				"provides and depends yet.",
+		},
 		/////////////////////////
 		// Version 3 specifics.//
 		/////////////////////////
--- a/cli/mender-artifact/write.go
+++ b/cli/mender-artifact/write.go
@@ -25,7 +25,36 @@ import (
 	"github.com/mendersoftware/mender-artifact/handlers"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
+	"io"
+	"io/ioutil"
 )
+
+func writeRootfsImageChecksum(c *cli.Context, typeInfo *artifact.TypeInfoV3) (err error) {
+	chk := artifact.NewWriterChecksum(ioutil.Discard)
+	payload, err := os.Open(c.String("file"))
+	if err != nil {
+		return cli.NewExitError(fmt.Sprintf("Failed to open the payload file: %q", c.String("file")), 1)
+	}
+	if _, err = io.Copy(chk, payload); err != nil {
+		return cli.NewExitError("Failed to generate the checksum for the payload", 1)
+	}
+	checksum := string(chk.Checksum())
+
+	Log.Debugf("Adding the `rootfs_image_checksum`: %q to Artifact provides", checksum)
+	if typeInfo == nil {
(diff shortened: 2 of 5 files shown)


oleorhagen:
@kacf, if you would just have a look, that would be nice. I think this is a fair approach. If you are happy, I will continue and add a quick unit test for it.

kacf:
Otherwise, the code looks good!

oleorhagen:
@kacf, please have another look.

The flag is added, however, what are your thoughts on it guiding this one addition, or should it also guard agains the user having added provides flags themselves?

cli/mender-artifact/write_test.go
```diff
@@ -399,3 +401,30 @@ func TestWriteRootfsArtifactDependsAndProvides(t *testing.T) {
...
+	require.NoError(t, tf.Close())
+	typeInfo := artifact.TypeInfoV3{}
+	set = flag.NewFlagSet("file", flag.ExitOnError)
+	set.String("file", tf.Name(), "foo")
+	err = writeRootfsImageChecksum(
+		cli.NewContext(nil, set, nil), &typeInfo)
+	assert.NoError(t, err)
+	require.NotNil(t, typeInfo.ArtifactProvides)
+	_, ok := (*typeInfo.ArtifactProvides)["rootfs_image_checksum"]
+	assert.True(t, ok)
```
kacf: This is probably something that can be inserted into an existing test, but I would also like to see one test for writing the whole artifact, and testing whether this info is there. I would be almost surprised if one of the existing tests aren't already complaining about the new artifact content. Have you run them all?

cli/mender-artifact/main.go
```diff
@@ -162,6 +162,12 @@ func getCliContext() *cli.App {
 			Usage: "Full path to the state script(s). You can specify multiple " +
 				"scripts providing this parameter multiple times.",
 		},
+		cli.BoolFlag{
+			Name: "no-provides",
```
kacf: > The flag is added, however, what are your thoughts on it guiding this one addition, or should it also guard agains the user having added provides flags themselves?

Hmm, good question. I think I'm leaning towards not disabling manually added ones, this seems like overreaching. IOW leave it as is.

However, it does bring into question the name of the flag, which is then a bit misleading. May I suggest renaming it to `no-checksum-provide`, to be more explicit?

cli/mender-artifact/write_test.go
oleorhagen: Oh yeh yeh, it breaks them spectacularly, lots of them, this is in no way ready for a PR. When I create these draft PR's it's usually for code discussion, not review ;)

cli/mender-artifact/write.go
```diff
@@ -25,8 +25,45 @@ import (
...
+	if _, err = io.Copy(chk, payload); err != nil {
+		return cli.NewExitError("Failed to generate the checksum for the payload", 1)
+	}
+	checksum := string(chk.Checksum())
+
+	Log.Debugf("Adding the `rootfs_image_checksum`: %q to Artifact provides", checksum)
+	if typeInfo == nil {
+		return errors.New("Type-info is unitialized")
+	}
+	if typeInfo.ArtifactProvides == nil {
```
oleorhagen: @kacf this is quite possibly the fugliest this in our code-base yet. I'm ashamed

oleorhagen: Magic is more than welcome here. Muggle hacking also for that matter.

oleorhagen: Ignore this fugliness, I got the shits and did: https://github.com/mendersoftware/mender-artifact/pull/246, so rebasing after it is merged.

kacf:
@oleorhagen: Let me know when this is ready, I'm not following up on every push.

oleorhagen:
@kacf have a look now

PR #247, opened 2019-12-20
Create artifact from device snapshot

alfrunes:
Using the ssh:// schema followed by a valid ssh address to a device with
mender, will attempt to receive a snapshot of currently running rootfs
from the device.

changelog: title


kacf:
@alfrunes: Let me know when you want me to look at this. I'm waiting since it says "WIP"

alfrunes:
Ok, @kacf; you can take a look at it now.
I'm not very comfortable with the implementation - it feels a bit "hacky", so I am open for suggestions.

cli/mender-artifact/write.go
```diff
@@ -460,3 +471,125 @@ func extractKeyValues(params []string) (*map[string]string, error) {
...
+	var written int64
+	for {
+		nr, err := stdout.Read(buf)
+		if err != nil {
+			cmd.Process.Kill()
+			return "", errors.Wrap(err,
+				"Error receiving snapshot from device")
+		}
+		nw, err := f.Write(buf[:nr])
+		if err == io.EOF {
```
kacf: This looks wrong. The last `err` came from `Write`, but you want to check the EOF condition of `Read`, don't you?

cli/mender-artifact/write.go
```diff
@@ -460,3 +471,125 @@ func extractKeyValues(params []string) (*map[string]string, error) {
...
+			break
+		} else if err != nil {
+			cmd.Process.Kill()
+			return "", errors.Wrap(err, "Error storing snapshot locally")
+		} else if nw < nr {
+			cmd.Process.Kill()
+			return "", io.ErrShortWrite
+		}
+		written += int64(nw)
+		fmt.Printf("\rReceiving snapshot from device... [%8s]", sizeStr(written))
```
kacf: Will this be intermingled with the output from the `mender` command on the device? Because stderr is connected to directly, so I assume that output will make it across.

cli/mender-artifact/write.go
```diff
@@ -460,3 +471,125 @@ func extractKeyValues(params []string) (*map[string]string, error) {
...
+	case <-ctx.Done():
+		// Context deadline exceeded
+		err = ctx.Err()
+	}
+
+	if err != nil {
+		cmd.Process.Kill()
+		return "", err
+	}
+	f, err := ioutil.TempFile("", "rootfs.ext4")
```
kacf: Btw, we need some kind of cleanup handling for this. I got failures for various reasons while playing with it, and these files then remain. They are often multi gigabyte files, so they should be cleaned up if there is a failure.

kacf:
After having played with it a bit, I also cannot find a better way of setting up the terminal in ssh and sudo. I think under the circumstances, you solution is the best we can do, despite being hacky. It is a quite clever solution though!

I have a couple of suggestions for improvement:

* Disable echo on the stdin file handle. I'm not sure how to do it in Go, but there is a [C example here](http://forums.codeguru.com/showthread.php?466009-Reading-from-stdin-(without-echo)&p=1786376#post1786376) (see the "termios" solution). This will help prevent password echoing for sudo, and it's unlikely that anything else will be needed for the stdin.

* I suggest the patch below. This will print all of the lines that are neither "aLL SET!" nor any of the payload data. Just in case an error appears that the user would otherwise not see.

   ```diff
   diff --git a/cli/mender-artifact/write.go b/cli/mender-artifact/write.go
   index 06a7249..b5cc399 100644
   --- a/cli/mender-artifact/write.go
   +++ b/cli/mender-artifact/write.go
   @@ -519,6 +519,7 @@ func getDeviceSnapshot(host string) (string, error) {
                           if strings.Contains(line, "aLL SET!") {
                                   break
                           }
   +                      os.Stdout.Write([]byte(line))
                   }      
                   errChan <- nil
           }()
   ```

cli/mender-artifact/write.go
alfrunes: That is indeed correct. Well spotted :+1:

cli/mender-artifact/write.go
alfrunes: I made it so that the progress bar will not print if the file descriptor is not a TTY device (in  `mender`). On a second thought I think it would be better to make it more explicit by adding a `--quiet` flag to the `snapshot dump` command.

alfrunes: For some reason ssh trims `\r` characters from the stderr stream, that's why I disabled the progress bar for non-tty devices in general.

kacf:
@alfrunes: Just FYI, I'm working on the SSH acceptance test as well.

kacf:
I'm [testing this](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/108112651) with [an acceptance test](https://github.com/mendersoftware/meta-mender/pull/893).

cli/mender-artifact/util/tty.go
```diff
@@ -0,0 +1,75 @@
...
+				sig == unix.SIGWINCH {
+				unix.Kill(os.Getpid(), sig.(unix.Signal))
+				continue
+			}
+			signal.Stop(sigChan)
+			errChan <- errors.Errorf("Received signal: %s",
+				unix.SignalName(sig.(unix.Signal)))
+			os.Remove(tmpFile)
+			if sig == unix.SIGCHLD {
+				sig = unix.SIGTERM
```
kacf: Why this assignment?

cli/mender-artifact/util/tty.go
```diff
@@ -0,0 +1,75 @@
...
+// Signal handler to re-enable tty echo on interrupt. The signal handler is
+// transparent with system default, and immedeately releases the channel and
+// calling the system sighandler after termios is set. To invoke it manually,
+// simply close the sigChan (make sure to call signal.Stop prior to closing).
+func EchoSigHandler(sigChan chan os.Signal, errChan chan error,
+	term *unix.Termios, tmpFile string) {
+	for {
+		sig, sigRecved := <-sigChan
+		// Restore Termios
+		unix.IoctlSetTermios(int(os.Stdin.Fd()), ioctlSetTermios, term)
```
kacf: Later on you may call `continue` which will return to the top of the loop, but by then the terminal will already be restored. Is it better to call this later once you know that you will not call `continue`?

cli/mender-artifact/util/tty.go
alfrunes: This wasn't supposed to be here... something leftover from earlier.

cli/mender-artifact/util/tty.go
alfrunes: Yeah, this is not the intention. Thanks for spotting this error, I moved the signal check before the ioctl statement.

kacf:
Alright, the backend tests are guaranteed to be unrelated (they passed before, and this is not even used in the backend). I'm merging this.

kacf:
We're not completely done though, because the we need the freeze fixes from https://github.com/mendersoftware/mender/pull/469. I could reproduce it locally, and have a [followup PR](https://github.com/mendersoftware/meta-mender/pull/897) which exposes this behavior, since it reuses the virtual machine for several tests (currently it's hanging because of the freeze bug).

PR #250, opened 2020-02-10
Accept suffix '.img' for mender-artifact modifiable images

lluiscampos:
Previously we were allowing only Yocto naming convention sdimg, uefiimg,
but it might come in handy to modify a .img image (i.e. converted
Raspbian).

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -26,7 +26,7 @@ import (
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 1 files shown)


lluiscampos:
I am unsure about this change deserving or not Changelog. What do you think?

lluiscampos:
Stopping integration pipeline for this PR.

lluiscampos:
> > I am unsure about this change deserving or not Changelog. What do you think?
> 
> I think it does. Small, but still a user facing change.

Ok, thanks.

> I'd let it run through the tests once though.

Sure. Running them now

PR #251, opened 2020-02-11
FIX: MergeArtifact[Depends|Provides] obeys json tags

alfrunes:
In the current state, the MergeArtifactDepends/Provides forces a value to unset struct values which conflicts with the json tags which has side-effects when attempting to match dependencies with provides.

### DIFF
--- a/artifact/metadata.go
+++ b/artifact/metadata.go
@@ -296,17 +296,38 @@ func (t TypeInfoDepends) Map() map[string]interface{} {
 }
 
 func NewTypeInfoDepends(m interface{}) (ti TypeInfoDepends, err error) {
+
+	const errMsgInvalidTypeFmt = "Invalid TypeInfo depends type: %T"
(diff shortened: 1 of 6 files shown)


artifact/metadata.go
```diff
@@ -346,17 +367,36 @@ func (t TypeInfoProvides) Map() map[string]interface{} {
 }
 
 func NewTypeInfoProvides(m interface{}) (ti TypeInfoProvides, err error) {
+
+	const errMsgInvalidTypeFmt = "Invalid TypeInfo provides type: %T"
```
oleorhagen: Question: Is it really necessary to pull these out when they are used in one place only?

alfrunes: It was done strictly out of formatting preference, it felt awkward to insert the write the string in-place and not violate the 80 character column width rule ;)
btw, this const is indirectly referenced two times as well.

PR #252, opened 2020-02-18
[pipeline] Set docker:dind image to a fixed version

lluiscampos:
So that we can workaround a temporary issue with latest docker:dind


lluiscampos:
Merging with no review as it is blocking work of others.

PR #256, opened 2020-03-05
Change ArtifactProvides.(type): map[string]interface{} -> map[string]string

alfrunes:
Only string properties are supported at this stage, let's not make it more complicated than it already is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.o
 *.a
 *.so
+*.mender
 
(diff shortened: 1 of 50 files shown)


alfrunes:
Sorry about all the amendments... I started a [new pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/123612449) with mendersoftware/meta-mender#940

PR #257, opened 2020-03-10
Bugfix (cli/snapshot): signal-loop on ignored signals

alfrunes:
If the forked ssh process received a signal which is ignored by default,
the signal is re-invoked using unix.Kill. This inevitably lead to an
invalid syscall instruction in case of SIGCHLD.

This commit makes SIGCHLD trigger terminal echo to be restored and
exits with an error. SIGURG and SIGWINCH are the only ignored signals.

### DIFF
--- a/cli/mender-artifact/write.go
+++ b/cli/mender-artifact/write.go
@@ -510,7 +510,7 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 	// established (password prompt is written to /dev/tty directly,
 	// and hence impossible to detect).
 	args = append(args, "sudo", "-S", "/bin/sh", "-c",
-		`'echo "`+sshInitMagic+`"; mender snapshot dump | cat'`)
+		`'echo "`+sshInitMagic+`" && mender snapshot dump | cat'`)
 
 	cmd := exec.Command("ssh", args...)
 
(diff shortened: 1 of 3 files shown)


lluiscampos:
That was fast @alfrunes! I think @kacf is best for the code review itself.

From my side, mention that the commit requires a changelog tag.

cli/mender-artifact/write.go
```diff
@@ -509,8 +509,8 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 	// First echo to stdout such that we know when ssh connection is
 	// established (password prompt is written to /dev/tty directly,
 	// and hence impossible to detect).
-	args = append(args, "sudo", "-S", "/bin/sh", "-c",
-		`'echo "`+sshInitMagic+`"; mender snapshot dump | cat'`)
+	args = append(args, "sudo", "-S", "-s",
+		`echo "`+sshInitMagic+`" && mender snapshot dump | cat`)
```
kacf: Probably somewhat theoretical, but I would argue that the original is better here. The old version guarantees that we will get the shell we want, whereas `-s` might not work if the user is using some weird shell.

lluiscampos:
> From my side, mention that the commit requires a changelog tag.

ping on this.

kacf:
And the ticket reference! See previous review.

alfrunes:
@lluiscampos @kacf - I fixed the commit message with a proper changelog tag and ticket id.

PR #258, opened 2020-03-11
Feature/dependabot

mzedel:


mzedel:
for now it is only meant to work for dependency notifications, but it seems to be where `go` is going for dependency management - this might make the transition easier once we fully go there...

PR #260, opened 2020-03-27
Add ability for artifact install to create directories

sbaxter-izo:
This PR adds the ability to create directories (and nested directories) to the `mender-artifact install` command, mimicking the `install` utility syntax:
`mender-artifact install -d <artifact>:/dirs/to/create`

Changelog: Title

Signed-off-by: Sam Baxter <sbaxter@izotope.com>

### DIFF
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -171,7 +171,7 @@ func Install(c *cli.Context) (err error) {
 			vdir, err := virtualImage.OpenDir(comp, privateKey, c.Args().First())
 			defer wclose(vdir)
 			if err != nil {
-				return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+				return cli.NewExitError(err, 1)
 			}
 
 			if err = vdir.Create(); err != nil {
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -137,7 +137,7 @@ func Copy(c *cli.Context) (err error) {
 	return nil
 }
 
-// Install installs a file from the host filesystem onto either
+// Install installs a file from the host filesystem or directory onto either
 // a mender artifact, or an sdimg.
 func Install(c *cli.Context) (err error) {
 	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
@@ -161,11 +161,24 @@ func Install(c *cli.Context) (err error) {
 	}
 	switch parseCLIOptions(c) {
 	case copyin:
+		directory := c.Bool("directory")
 		var perm os.FileMode
-		if c.Int("mode") == 0 {
+		if c.Int("mode") == 0 && !directory {
 			return cli.NewExitError("File permissions needs to be set, if you are simply copying, the cp command should fit your needs", 1)
 		}
 		perm = os.FileMode(c.Int("mode"))
+		if directory {
+			vdir, err := virtualImage.OpenDir(comp, privateKey, c.Args().First())
+			defer wclose(vdir)
+			if err != nil {
+				return cli.NewExitError(fmt.Sprintf("%v", err), 1)
+			}
+
+			if err = vdir.Create(); err != nil {
+				return cli.NewExitError(err, 1)
+			}
+			return nil
+		}
 		f, err := os.Open(c.Args().First())
 		defer f.Close()
 		if err != nil {
@@ -198,7 +211,7 @@ func Install(c *cli.Context) (err error) {
 	case parseError:
 		return cli.NewExitError("No artifact or sdimg provided", 1)
 	case argerror:
-		return cli.NewExitError(fmt.Sprintf("got %d arguments, wants two", c.NArg()), 1)
+		return cli.NewExitError(fmt.Sprintf("Wrong number of arguments, got %d", c.NArg()), 1)
 	default:
 		return cli.NewExitError("Unrecognized parse error", 1)
 	}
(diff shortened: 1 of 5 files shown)


cli/mender-artifact/debugfs.go
```diff
@@ -73,6 +73,22 @@ func debugfsCopyFile(file, image string) (ret string, err error) {
...
 }
 
+func debugfsMakeDir(imageFile, image string) (err error) {
+	// Remove the `/` suffix if present, as debugfs does not play nice with it.
+	imageFile = strings.TrimRight(imageFile, "/")
+	// Recursively create parent directories if they do not exist
+	dir, _ := filepath.Split(imageFile)
+	_, err = os.Stat(dir)
+	if dir != "" && os.IsNotExist(err) {
+		debugfsMakeDir(dir, image)
```
kacf: I don't quite follow this code. Isn't `imageFile` the path inside of the image? So why is `dir` being tested using `os.Stat` on the host filesystem?

cli/mender-artifact/main.go
```diff
@@ -395,8 +395,8 @@ func getCliContext() *cli.App {
 
 	install := cli.Command{
 		Name:        "install",
-		Usage:       "install -m <permissions> <hostfile> [artifact|sdimg|uefiimg]:<filepath>",
-		Description: "Installs a file from the host filesystem to the artifact or sdimg.",
+		Usage:       "install -m <permissions> <hostfile> [artifact|sdimg|uefiimg]:<filepath> or install -d [artifact|sdimg|uefiimg]:<directory>",
+		Description: "Installs a file or directory from the host filesystem to the artifact or sdimg.",
```
kacf: Suggest rephrasing to `Installs a directory, or a file from the host filesystem, to the artifact or sdimg`.

cli/mender-artifact/main.go
```diff
@@ -405,6 +405,10 @@ func getCliContext() *cli.App {
 			Name:  "mode, m",
 			Usage: "Set the permission bits in the file",
 		},
+		cli.BoolFlag{
+			Name:  "directory, d",
+			Usage: "Install a directory into an artifact",
```
kacf: I would rephrase as `Create a directory inside an artifact`.

cli/mender-artifact/debugfs.go
sbaxter-izo: Good catch! The UNIX `install -d` silently does nothing if the directory already exists, so I'll mirror that behavior.

cli/mender-artifact/main.go
sbaxter-izo: Fixed the phrasing

cli/mender-artifact/main.go
sbaxter-izo: Fixed the phrasing.

cli/mender-artifact/debugfs.go
sbaxter-izo: I removed this `os.Stat` logic and replaced it with a `debugfs stat` check to simply return if the directory already exists.

cli/mender-artifact/copy.go
```diff
@@ -161,11 +161,24 @@ func Install(c *cli.Context) (err error) {
...
-		if c.Int("mode") == 0 {
+		if c.Int("mode") == 0 && !directory {
 			return cli.NewExitError("File permissions needs to be set, if you are simply copying, the cp command should fit your needs", 1)
 		}
 		perm = os.FileMode(c.Int("mode"))
+		if directory {
+			vdir, err := virtualImage.OpenDir(comp, privateKey, c.Args().First())
+			defer wclose(vdir)
+			if err != nil {
+				return cli.NewExitError(fmt.Sprintf("%v", err), 1)
```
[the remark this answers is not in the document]
sbaxter-izo: Removed `fmt.Sprintf`

sbaxter-izo:
Thanks for the review!

> Do I understand correctly that it works more or less like the Unix `install -d` tool? This means that the directory does not necessarily have to exist on the build host filesystem, correct?

Yep, that is the intention and I've resoled the requested changes. Ready for another look.

cli/mender-artifact/debugfs.go
```diff
@@ -74,12 +74,16 @@ func debugfsCopyFile(file, image string) (ret string, err error) {
 }
 
 func debugfsMakeDir(imageFile, image string) (err error) {
+	_, err = debugfsExecuteCommand(fmt.Sprintf("stat %s", imageFile), image)
```
kacf, line 77: Should this also go after the `TrimRight`?

sbaxter-izo: `debugfs stat` doesn't mind the trailing `/`, but `mkdir` does. I've clarified the comment to say it is needed for the `mkdir` command

sbaxter-izo:
Ah, darn, I forgot that. Fixed now.

Thanks for the review! Every time I've interacted with the mender community I've received great support, happy to contribute.

sbaxter-izo:
@kacf Is there anything with the failing checks that I need to resolve so that this can get merged?

I'm also curious how/when this upstream work gets integrated into the `meta-mender` upstream. We're upgrading from an older poky to latest zeus. I've noticed there doesn't seem to be zeus support in `meta-mender` upstream yet either, but I'm wondering if this change will make it in there once zeus support is complete.

Thanks again!

kacf:
It appears that the current failure is due to some unrelated network timeout, so I'm rerunning that one.

This will end up both in zeus and in older branches, don't worry about that! zeus doesn't have its own branch yet because we've had a lot of integration problems with it, but it is being worked on.

sbaxter-izo:
Great to hear!

kacf:
Thanks @sbaxter-izo!

PR #268, opened 2020-04-27
MEN-3465: Disallow empy 'device-type' meta-data in Artifacts

oleorhagen:
This is two-sprung:

* Disallow writing of module images with no `device-type` set
* Fail parsing Artifact payloads with no 'device-types'

Ticket: https://tracker.mender.io/browse/MEN-3465


cli/mender-artifact/main.go
```diff
@@ -204,7 +204,7 @@ func getCliContext() *cli.App {
 	writeModuleCommand.Flags = []cli.Flag{
 		cli.StringSliceFlag{
 			Name: "device-type, t",
-			Usage: "Type of device(s) supported by the Artifact. You can specify multiple " +
+			Usage: "[Required] Type of device(s) supported by the Artifact. You can specify multiple " +
```
alfrunes, line 207: I believe there is a `Required` property in the flag struct that will check that a value is present after parsing.

artifact/metadata.go
```diff
@@ -263,6 +263,31 @@ type ArtifactDepends struct {
...
 
+type Error struct {
+	Err string
+}
+
+func (e *Error) Error() string {
+	return e.Err
+}
+
+var ErrCompatibleDevices *Error = &Error{"ArtifactDepends: Required field 'CompatibleDevices' not found"}
```
alfrunes, line 274: Why not use standard error? `... = errors.New("ArtifactDepends...`

oleorhagen: Fair point. Initially I thought about adding more information to the error.

Removing.

cli/mender-artifact/main.go
oleorhagen: I couldn't find it in the  version of the cli we're using. Did I miss it?

I did double check it now also

artifact/metadata.go
kacf: +1

cli/mender-artifact/main.go
kacf: If we are adding a `[Required]` tag here we should be consistent, and there are several other arguments which are also required. You should double check the list by manual checking, but I know at least `--type` for module-image and `--file` for rootfs-image are required, and `--name` is required for both of them.

cli/mender-artifact/read.go
```diff
@@ -81,6 +81,7 @@ func readArtifact(c *cli.Context) error {
 	err = ar.ReadArtifact()
 	if err != nil {
 		return cli.NewExitError(err.Error(), 0)
+		return cli.NewExitError(err.Error(), 1)
```
kacf, line 87: I take it you wanted to remove the other one? 🙂

artifact/metadata.go
```diff
@@ -263,6 +263,31 @@ type ArtifactDepends struct {
...
+
+func (a *ArtifactDepends) UnmarshalJSON(b []byte) error {
+	type Alias ArtifactDepends // Same fields, no inherited UnmarshalJSON method
+	buf := &Alias{}
+	if err := json.Unmarshal(b, buf); err != nil {
+		return err
+	}
+	if len(buf.CompatibleDevices) == 0 {
+		return ErrCompatibleDevices
+	}
```
kacf, line 284: Will this work for v2 artifacts? I'm not sure, but they don't store `CompatibleDevices` in `ArtifactDepends`.

oleorhagen: Well spotted! Ur right of course. This has to be added.

oleorhagen: But also, `update-module`s are not supported in `v2`. I will still add the check though.

cli/mender-artifact/read.go
oleorhagen: I will fix this  up last, so that I don't have to force push during review.

cli/mender-artifact/main.go
oleorhagen: I see, might as well standardize this then.

This means I'll update the `urfave/cli` dependency also.

PR #269, opened 2020-04-28
Moved Mender-Artifact to go mod

oleorhagen:
At first I thought it was gonna be a five minute job, but turns out go.mod had a few tricks up its sleeve.

For one, it changes the permissions on the files it vendors, so that the mendertesting scripts had to be first vendored, and then I explicitly changed the scripts permissions back to executable.

Also, first, I wanted to make do, without the `./vendor` folder completely, but this requires some deeper changes to our license checks, which I did not want to do at this moment in time. If we want, this can be accomplished at a later time, if we so desire.

- O


lluiscampos:
@oleorhagen Remember that we agreed in doing this after 2.4 release is branched out. So please don't merge it just yet.

oleorhagen:
> @oleorhagen Remember that we agreed in doing this after 2.4 release is branched out. So please don't merge it just yet.

I know, had a discussion with @kacf today though, where the basis is that some dependencies (the ones with v2 extensions) are a pain to work with when not using go modules. Therefore we decided to give it a run in `mender-artifact`. The change should be pretty much transparent.

Thus unless you have some strong objections, I think we are going to go ahead with this for this repo only.

kacf:
So let me see if I understand the changes correctly: You add the `go.mod` file, and then an extra file, `vendor/modules.txt` (I assume this is an optional file) which mirrors the first one. Then the vendor folder is synchronized to match that file?

> I know, had a discussion with @kacf today though, where the basis is that some dependencies (the ones with v2 extensions) are a pain to work with when not using go modules. Therefore we decided to give it a run in `mender-artifact`. The change should be pretty much transparent.
> 
> Thus unless you have some strong objections, I think we are going to go ahead with this for this repo only.

Yes, it's needed for a change on the current sprint. Not strongly needed though (it's only the ["Required" tag](https://github.com/mendersoftware/mender-artifact/pull/268), right?), if this becomes too problematic we could still back out.

oleorhagen:
> So let me see if I understand the changes correctly: You add the `go.mod` file, and then an extra file, `vendor/modules.txt` (I assume this is an optional file) which mirrors the first one. Then the vendor folder is synchronized to match that file?
> 
> > I know, had a discussion with @kacf today though, where the basis is that some dependencies (the ones with v2 extensions) are a pain to work with when not using go modules. Therefore we decided to give it a run in `mender-artifact`. The change should be pretty much transparent.
> > Thus unless you have some strong objections, I think we are going to go ahead with this for this repo only.
> 
> Yes, it's needed for a change on the current sprint. Not strongly needed though (it's only the ["Required" tag](https://github.com/mendersoftware/mender-artifact/pull/268), right?), if this becomes too problematic we could still back out.

Exactly. So nice to have. Not required.

lluiscampos:
> > Yes, it's needed for a change on the current sprint. Not strongly needed though (it's only the ["Required" tag](https://github.com/mendersoftware/mender-artifact/pull/268), right?), if this becomes too problematic we could still back out.
> 
> Exactly. So nice to have. Not required.

All right. I trust you on this. Green light from my side :+1:

go.mod
```diff
@@ -1,14 +1,27 @@
 module github.com/mendersoftware/mender-artifact
 
+go 1.14
```
mzedel, line 3: this should align with: https://github.com/mendersoftware/mender-artifact/blob/master/Dockerfile#L1 as it's built within docker... I'd suggesting removing it here, upgrading the base image to use a 1.14 based one and verifying the go mod upgrades in the image...

oleorhagen: Yes, sir!

Just to be clear that we are in sync here.

You are suggesting removing the `1.14` version in go.mod?

Or simply upgrading it to `1.14` in the Dockerfile?

mzedel: both 😄 - so that we at best have only place to set the go version, while the rest just picks it up... otherwise we could end up having some packages installed that require 1.14 and then fail because the image it's built in is just 1.11...

oleorhagen: I don't see how I can add this in one place though, unless I'm missing something obvious?

If not, I will upgrade the `Dockerfile` to `1.14`

mzedel: During the build (which takes place according to the `Dockerfile`) the environment `go.mod` will be executed in, has a go version set - so the setting in `go.mod` shouldn't be required, since the environment already dictates it.

oleorhagen: The structure of the `go.mod` file sort of requires this.

It is a 'minimum' tag for building this binary, sort of. Although I don't think it's much enforced yet. Older version will still try to build, and seems like they succeed mostly too.

Still, I can't remove that line.

Does that make sense?

mzedel: ah sorry so it's more like https://github.com/mapbox/mapbox-gl-js/blob/master/package.json#L13, then I just completely misunderstood that line... 🙈- having the version aligned with the `Dockerfile` should still be good to have I guess...

oleorhagen: Yes, seems exactly like that :)

Yes, did upgrade the go versions in the Dockerfiles to align :racehorse:

oleorhagen:
Pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/pipelines/140978711

PR #272, opened 2020-05-25
MEN-3580: Fix creating artifact from snapshot terminating on SIGCHLD

alfrunes:
changelog: none

### DIFF
--- a/cli/mender-artifact/util/tty.go
+++ b/cli/mender-artifact/util/tty.go
@@ -60,10 +60,16 @@ func EchoSigHandler(
 		unix.IoctlSetTermios(int(os.Stdin.Fd()), ioctlSetTermios, term)
 		if sigRecved {
 			signal.Stop(sigChan)
-			errChan <- errors.Errorf("Received signal: %s",
-				unix.SignalName(sig.(unix.Signal)))
-			// Relay signal to default handler
-			unix.Kill(os.Getpid(), sig.(unix.Signal))
+			switch sig {
+			case unix.SIGCHLD:
+				// SIGCHLD is expected when ssh terminates.
+				errChan <- nil
+			default:
+				errChan <- errors.Errorf("Received signal: %s",
+					unix.SignalName(sig.(unix.Signal)))
+				// Relay signal to default handler
+				unix.Kill(os.Getpid(), sig.(unix.Signal))
+			}
 		} else {
 			errChan <- nil
 		}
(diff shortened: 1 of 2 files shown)


cli/mender-artifact/util/tty.go
```diff
@@ -60,10 +60,18 @@ func EchoSigHandler(
...
+				// SIGCHLD is expected when ssh terminates.
+				errChan <- nil
+			default:
+				errChan <- errors.Errorf("Received signal: %s",
+					unix.SignalName(sig.(unix.Signal)))
+				// Relay signal to default handler
+				unix.Kill(os.Getpid(), sig.(unix.Signal))
+			}
+			if sig == unix.SIGCHLD {
+			}
```
lluiscampos, line 74: This can be cleaned up.

kacf:
> Question: how comes that we didn't catch this in an acceptance test? I think we should extend one the tests (or create a new one) to test this scenario and make sure we don't break it again.

Indeed, I'm very curious why this wasn't caught. This exact scenario is tested in the acceptance tests. I can take a look at the test today.

kacf:
Don't merge it yet, btw, I want to try fixing the test first.

kacf:
When testing this manually, the SIGCHLD message is gone, but I still get an error:

```
................................ 100% 229376 KiB
can not create artifact file
```

alfrunes:
> When testing this manually, the SIGCHLD message is gone, but I still get an error:
> 
> ```
> ................................ 100% 229376 KiB
> can not create artifact file
> ```

This error message is returned when `os.Create` returns an error (it does not print the actual error). I was not able to reproduce this error, but I appended the underlying error to the error message. Can you try again and see if the error makes sense now?

kacf:
> > When testing this manually, the SIGCHLD message is gone, but I still get an error:
> > ```
> > ................................ 100% 229376 KiB
> > can not create artifact file
> > ```
> 
> This error message is returned when `os.Create` returns an error (it does not print the actual error). I was not able to reproduce this error, but I appended the underlying error to the error message. @kacf; Can you try again and see if the error makes sense now?

It helped! Made me realize that I had somehow made the existing file root-owned, and was trying to overwrite it (I was messing with Docker containers earlier). So this was a pilot error. Still, having more verbose messaging is always good, so I vote for keeping that!

I'm still looking into why the test passed, stay tuned!

kacf:
@alfrunes: I give up. The test passes when the input/output is a log file, and fails when it's a terminal. Since running with a terminal is not practical in Gitlab, I tried to find out why this makes a difference to SIGCHLD, but I came up empty handed. The next step would be to step through the code line by line to see exactly what causes SIGCHLD to reported only in the terminal case, but I think this is too much effort just for a test. At least we are guaranteed to catch it during releases, since the snapshotting is part of the Quick Start guide.

So I think this can be merged as it is. Just hit the merge button if you agree!

PR #279, opened 2020-06-10
"dependabot update"

mzedel:
"migrated to stable dependabot release"

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,14 @@
+version: 2
+updates:
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
+  - commit-message:
+      prefix: "Changelog:All"
+    directory: /
+    package-ecosystem: docker
+    schedule:
+      interval: weekly
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,8 +1,0 @@
-version: 1
-update_configs:
-  - package_manager: 'go:modules'
-    directory: '/'
-    update_schedule: 'weekly'
-  - package_manager: 'docker'
-    directory: '/'
-    update_schedule: 'weekly'


.github/dependabot.yml
```diff
@@ -0,0 +1,14 @@
+version: 2
+updates:
+  - commit-message:
+      prefix: "Updated dependencies\nChangelog: Commit\n"
+    directory: /
+    package-ecosystem: gomod
+    schedule:
+      interval: weekly
+  - commit-message:
+      prefix: "Updated dependencies\nChangelog: Commit\n"
```
lluiscampos, line 10: @mzedel I think with the "Updated dependencies" part it might get too verbose. The commit already has a title (which is included when using Changelog: Commit") which already indicates that a given dependency is updated. Furthermore this line would be in the middle of the Changelog entry...

Other than that, good work with these PRs!

lluiscampos: (if change this, we shall do it in all repos; but I am not going to comment in the rest)

mzedel: Ah ok, then using [this commit](https://github.com/mendersoftware/gui/commit/7f96f8a0f8b33764f5bf4fe11b99be25eef0079a) as an example the entire commit would be pretty verbose as well, What would you say about just using `Changelog: Title`?

kacf: I think `Changelog: Commit` is really useful here, since it points to the changelogs of the other projects. It's still short, but very informational. This is a textbook example of how a changelog entry should be, IMHO.

mzedel: @kacf what would be your opinion about this?

mzedel: after merging the config in the integration repo => https://github.com/mendersoftware/integration/network/updates shows that there is a length limitation... unfortunately there is no linter yet for the stable version ☹️

lluiscampos: @mzedel I see that you have now reworked the different PRs to use `Changelog:All` :+1:

I have created [QA-196](https://tracker.mender.io/browse/QA-196) accordingly; I will work on it in the upcoming days, but note that the task it is not a blocker to merge these set of PRs.

The only blocker for these PRs is to do a "release" of `mendertesting` repo as a go module (which I personally don't know how to do). Doing such release will make the dependabot to update the dependency in all repos, hence getting the latest changes in `check_commits` which will make the PRs pass.

mzedel: Thanks for the adjusted check @lluiscampos - it has already worked in https://gitlab.com/Northern.tech/Mender/integration/-/jobs/592537988 🎉 
regarding the release: apparently you need a tag to bind the release to, and then we can create a release through [the github ui](https://github.com/mendersoftware/mendertesting/releases/new) or via the [api](https://developer.github.com/v3/repos/releases/#create-a-release)... I wanted to piggyback on the [dependabot PR in mendertesting](https://github.com/mendersoftware/mendertesting/pull/83) but could only get the tag in once the commit is merged 😅

PR #288, opened 2020-09-03
PoC: Dynamically load mender-artifact update module

oleorhagen:
# PoC - Plugin functionality

This is a PoC of the concept of dynamically loading functionality into `mender-artifact`.

The idea is to collect the somewhat disparate functionality present in our `generate-<update-module>` scripts. And pull them all back into `mender-artifact`, which is where they belong imo.

This way, every `generator` script we now have, can be collected into `mender-artifact`, which:
* increases the visibility and discover-ability of the update-modules available.
* gives all update-module functionality auto-completion, and full help-text directly in the tool. No `-- <discover-the-mender-artifact-flags-yourself>` prints.
* gives us the opportunity to distribute the `delta` generator as a plugin to `mender-artifact`, as compared to a separate script.
* It removes the need for hacks, such as the `dump` command, as plugins can work directly with the existing `mender-artifact` functionality. No need to serialize, and deserialize data.

There are some cons which should be noted, as they can come back and bite us, if not handled properly. (See [cons](#cons))

### Old

```
$ Download the specific generator tool
$ <run-script> -- <non-discoverable-mender-artifact-flags>
```

### New

```
$ mender-artifact generate -h
NAME:
   mender-artifact generate - Generates special Artifacts

USAGE:
   mender-artifact generate command [command options] [arguments...]

COMMANDS:
   delta  Writes Mender artifact for an delta update module
   k8s    Writes Mender artifact for an k8s update module

OPTIONS:
   --help, -h  show help
```
```
$ mender-artifact git:(pluginpoc) ✗ ./mender-artifact generate k8s -h
NAME:
   mender-artifact generate k8s - Writes Mender artifact for an k8s update module

USAGE:
   Writes a k8s Mender artifact that will be used by the k8s update module. 

REQUIRED ARGUMENTS:
   --artifact-name value, -n value  Name of the artifact
   --device-type value, -t value    Type of device(s) supported by the Artifact. You can specify multiple compatible devices providing this parameter multiple times.
   --manifestos value, --ma value   Path(s) to the Kubernetes manifesto(s)

OPTIONS:
   --artifact-name-depends value, -N value  Sets the name(s) of the artifact(s) which this update depends upon
   --augment-depends KEY:VALUE              Generic KEY:VALUE which is added to the augmented type-info -> artifact_depends section. Can be given multiple times
   --augment-file FILE                      Include FILE in payload in the augment section. Can be given more than once.
   --augment-meta-data FILE                 The meta-data JSON FILE for this payload, for the augmented section
   --augment-provides KEY:VALUE             Generic KEY:VALUE which is added to the augmented type-info -> artifact_provides section. Can be given multiple times
   --augment-type value                     Type of augmented payload. This is the same as the name of the update module
   --compression value                      Compression to use for data and header inside the artifact, currently supports: none, gzip, lzma.
   --depends KEY:VALUE, -d KEY:VALUE        Generic KEY:VALUE which is added to the type-info -> artifact_depends section. Can be given multiple times
   --depends-groups value, -G value         The group(s) the artifact depends on
   --file FILE, -f FILE                     Include FILE in payload. Can be given more than once.
   --key value, -k value                    Full path to the private key that will be used to sign the artifact.
   --meta-data FILE, -m FILE                The meta-data JSON FILE for this payload
   --output-path value, -o value            Full path to output artifact file.
   --provides KEY:VALUE, -p KEY:VALUE       Generic KEY:VALUE which is added to the type-info -> artifact_provides section. Can be given multiple times
   --provides-group value, -g value         The group the artifact provides
   --script value, -s value                 Full path to the state script(s). You can specify multiple scripts providing this parameter multiple times.
   --version value, -v value                Version of the artifact. (default: 3)
```

Which perfectly matches the functionality already present in `mender-artifact`, except this can be extended to any other update-module at runtime by creating a plugin.

Or, if autocompletion is installed, you can <tab> yourself to find the options available interactively. 
   

## Comments on the CLI command tree chosen

In general, I think there are two ways to go about the CLI here:

1. In this PoC a new CLI option `generate`, will show up in the first submenu, if a plugin is found. (which is what this PR does)
2. Is to add them under the existing `write module-image` command, perhaps cleaner?

> I'd like to discuss which approach is favoured here.

For this commit, that is a dummy `k8s` module, which simply validates a manifesto file. Extending it to full functionality is easy though.

To test this locally, do:
```
$ sudo make build build-plugins install-plugins
$ ./mender-artifact generate -h
# ./mender-artifact generate k8s --manifestos <some-manifesto-or-dummy-file> -n foo -t bar
```

For extra visibility, have the autocompletion functionality installed (optional, but recommended):

```
$ sudo make install-autocomplete-scripts
```

## Cons
This relies on the `plugin` module in go, to dynamically load the plugin at runtime. This does have some serious drawbacks, which we should be aware of.

The plugin and the `mender-artifact` binary, has to be built with the same version of `go`, and the same version of the packages they both share (notably `urfave/cli`).  In general I think this can be solved, simply by having them as a part of the release process, so that they are built and distributed in and from the same environment. 

For someone not happy with `go plugins`, see [reddit](https://www.reddit.com/r/golang/comments/b6h8qq/is_anyone_actually_using_go_plugins/)

I guess this can be solved through always building with our Docker build files?

### DIFF
--- a/cli/mender-artifact/main.go
+++ b/cli/mender-artifact/main.go
@@ -16,12 +16,14 @@ package main
 
 import (
 	"fmt"
+	"io/ioutil"
 	"os"
-	"strings"
+	"path"
+	"plugin"
 	"sort"
+	"strings"
 
 	"github.com/mendersoftware/mender-artifact/artifact"
-
 	"github.com/urfave/cli"
 )
 
@@ -38,6 +40,10 @@ const (
 
 // Version of the mender-artifact CLI tool
 var Version = "unknown"
+var pluginDirs = [...]string{
+	os.Getenv("HOME"),
+	"/etc/mender-artifact/",
+}
 
 // LatestFormatVersion is the latest version of the format, which is
 // also what we default to.
@@ -97,8 +103,8 @@ func getCliContext() *cli.App {
 	// Common Artifact flags
 	//
 	artifactName := cli.StringFlag{
-		Name:  "artifact-name, n",
-		Usage: "Name of the artifact",
+		Name:     "artifact-name, n",
+		Usage:    "Name of the artifact",
 		Required: true,
 	}
 	artifactNameDepends := cli.StringSliceFlag{
@@ -210,7 +216,7 @@ func getCliContext() *cli.App {
 
 	writeModuleCommand.CustomHelpTemplate = CustomSubcommandHelpTemplate
 
-	writeModuleCommand.Flags = []cli.Flag{
+	tmpFlagHolder := []cli.Flag{
 		cli.StringSliceFlag{
 			Name: "device-type, t",
 			Usage: "Type of device(s) supported by the Artifact. You can specify multiple " +
@@ -239,11 +245,12 @@ func getCliContext() *cli.App {
 		artifactNameDepends,
 		artifactProvidesGroup,
 		artifactDependsGroups,
-		cli.StringFlag{
-			Name:  "type, T",
-			Usage: "Type of payload. This is the same as the name of the update module",
-			Required: true,
-		},
+		// Replaced with dynamically loaded sub-commands
+		// cli.StringFlag{
+		// 	Name:     "type, T",
+		// 	Usage:    "Type of payload. This is the same as the name of the update module",
+		// 	Required: true,
+		// },
 		payloadProvides,
 		payloadDepends,
 		payloadMetaData,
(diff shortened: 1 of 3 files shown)


kacf:
I like very much idea! But the question is if this is a good idea with the tools we have available, see below.

> The plugin and the `mender-artifact` binary, has to be built with the same version of `go`, and the same version of the packages they both share (notably `urfave/cli`). In general I think this can be solved, simply by having them as a part of the release process, so that they are built and distributed in and from the same environment.

Darn, this is a quite serious restriction. I would have expected better from Go, TBH. I suppose it is somewhat solvable for our own modules, but what about other people's modules? Also, we don't want to re-release mender-binary-delta every time we upgrade Go version.

I'm worried that this is going to bite us if every (or almost every) upgrade of mender-artifact requires an upgrade of all other modules as well. The idea was that these would be pretty independent. Do you see any other way around it?

oleorhagen:
> Darn, this is a quite serious restriction. I would have expected better from Go, TBH. I suppose it is somewhat solvable for our own modules, but what about other people's modules?

Could be solved, as we build all our source with the Docker containers, I guess. But adds a bit of an entry barrier, but this is not the deal-breaker imo.

> Also, we don't want to re-release mender-binary-delta every time we upgrade Go version.

This is the more pressing concern to me. 

And might actually be the deal-breaker here. It is a little retarded, if we have to do this.

> 
> I'm worried that this is going to bite us if every (or almost every) upgrade of mender-artifact requires an upgrade of all other modules as well. 

Yup, at least for user-created custom ones. However, I do agree that the cost here outweighs the benefits.

> The idea was that these would be pretty independent. Do you see any other way around it?

I will keep looking. Seems hashicorp has some `go-plugin` package using `gRpc`. 

> I like very much this idea

Me 2. *I think this could add great value to our users*. So this was mostly to highlight the use-case. Which it seems you agree with. I'm therefore going to continue on this journey, and see if I can get rid of some of these pretty serious drawbacks!

cli/mender-artifact/main.go
```diff
@@ -291,7 +345,7 @@ func getCliContext() *cli.App {
 	validate := cli.Command{
 		Name:        "validate",
 		Usage:       "Validates artifact file.",
-		Category: "Artifact creation and validation",
+		Category:    "Artifact creation and validation",
```
lluiscampos, line 348: All these spaces fixes you can actually commit in a separate PR :)

lluiscampos (commented):
As informally commented, this is a great idea and I agree with you both that it brings high value to our customers. I had a look a the code now (my original "informal" comment was without code) and it seems very intuitive to write these plugins.

Of course I also agree with the limitations, and I encourage you to invest more value add time here @oleorhagen .

Great work!

oleorhagen:
I am closing this as of now:

My research this far has looked into alternatives such as:
* Yaegi
* go-plugin
Neither of which bring more value than they add in complexity, in terms of how I wanted this to work.

I am therefore closing this as of now, if a new go release contains improvements for dynamically loaded code and a stable ABI, then I will bring this back to life.

For now, this is is too much effort considering the extra amount of versioning it requires

:'(

kacf:
:sob:

PR #291, opened 2020-09-23
MEN-3076,MEN-3077: set software version in {rootfs,module}-image

tranchitella:
This PR also includes MEN-3482: Change rootfs_image_checksum over to use namespaced provides.

Depends on:
* https://github.com/mendersoftware/integration/pull/1016
* https://github.com/mendersoftware/meta-mender/pull/1093
* https://github.com/mendersoftware/meta-mender/pull/1096

### DIFF
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -330,13 +330,13 @@ func repack(comp artifact.Compressor, ua *unpackedArtifact, to io.Writer, key []
 		aWriter = awriter.NewWriterSigned(to, comp, artifact.NewSigner(key))
 	}
 
-	// for rootfs-images: Update rootfs_image_checksum provide if there is one.
-	_, hasChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs_image_checksum"]
+	// for rootfs-images: Update rootfs-image.checksum provide if there is one.
+	_, hasChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs-image.checksum"]
 	if ua.writeArgs.TypeInfoV3.Type == "rootfs-image" && hasChecksumProvide {
 		if len(ua.files) != 1 {
 			return errors.New("Only rootfs-image Artifacts with one file are supported")
 		}
-		err := writeRootfsImageChecksum(ua.files[0], ua.writeArgs.TypeInfoV3)
+		err := writeRootfsImageChecksum(ua.files[0], ua.writeArgs.TypeInfoV3, false)
 		if err != nil {
 			return err
 		}
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -332,11 +332,13 @@ func repack(comp artifact.Compressor, ua *unpackedArtifact, to io.Writer, key []
 
 	// for rootfs-images: Update rootfs-image.checksum provide if there is one.
 	_, hasChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs-image.checksum"]
-	if ua.writeArgs.TypeInfoV3.Type == "rootfs-image" && hasChecksumProvide {
+	// for rootfs-images: Update legacy rootfs_image_checksum provide if there is one.
+	_, hasLegacyChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs_image_checksum"]
+	if ua.writeArgs.TypeInfoV3.Type == "rootfs-image" && (hasChecksumProvide || hasLegacyChecksumProvide) {
 		if len(ua.files) != 1 {
 			return errors.New("Only rootfs-image Artifacts with one file are supported")
 		}
-		err := writeRootfsImageChecksum(ua.files[0], ua.writeArgs.TypeInfoV3, false)
+		err := writeRootfsImageChecksum(ua.files[0], ua.writeArgs.TypeInfoV3, hasLegacyChecksumProvide)
 		if err != nil {
 			return err
 		}
--- a/cli/mender-artifact/main.go
+++ b/cli/mender-artifact/main.go
@@ -17,8 +17,8 @@ package main
 import (
 	"fmt"
 	"os"
-	"strings"
 	"sort"
+	"strings"
 
 	"github.com/mendersoftware/mender-artifact/artifact"
 
@@ -34,6 +34,13 @@ const (
 	errArtifactInvalid
 	errArtifactUnsupportedFeature
 	errSystemError
+)
+
+const (
+	noDefaultSoftwareVersionFlag = "no-default-software-version"
+	softwareNameFlag             = "software-name"
+	softwareVersionFlag          = "software-version"
+	softwareFilesystemFlag       = "software-filesystem"
 )
 
 // Version of the mender-artifact CLI tool
@@ -97,8 +104,8 @@ func getCliContext() *cli.App {
 	// Common Artifact flags
 	//
 	artifactName := cli.StringFlag{
-		Name:  "artifact-name, n",
-		Usage: "Name of the artifact",
+		Name:     "artifact-name, n",
+		Usage:    "Name of the artifact",
 		Required: true,
 	}
 	artifactNameDepends := cli.StringSliceFlag{
@@ -112,6 +119,24 @@ func getCliContext() *cli.App {
 	artifactDependsGroups := cli.StringSliceFlag{
 		Name:  "depends-groups, G",
 		Usage: "The group(s) the artifact depends on",
+	}
+
+	// Common Software Version flags
+	softwareVersionNoDefault := cli.BoolFlag{
+		Name:  noDefaultSoftwareVersionFlag,
+		Usage: "Disable the software version field for compatibility with old clients",
+	}
+	softwareVersionName := cli.StringFlag{
+		Name:  softwareNameFlag,
+		Usage: "Name of the key to store the software version: rootfs-image.NAME.version, instead of rootfs-image.version",
+	}
+	softwareVersionValue := cli.StringFlag{
+		Name:  softwareVersionFlag,
+		Usage: "Value for the software version, defaults to the name of the artifact",
+	}
+	softwareFilesystem := cli.StringFlag{
+		Name:  softwareFilesystemFlag,
+		Usage: "If specified, is used instead of rootfs-image",
 	}
 
 	//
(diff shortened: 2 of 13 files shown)


cli/mender-artifact/main.go
```diff
@@ -272,12 +296,18 @@ func getCliContext() *cli.App {
 			Usage: "Include `FILE` in payload in the augment section. Can be given more than once.",
 		},
 		compressionFlag,
+		//////////////////////
+		// Sotware versions //
+		//////////////////////
+		softwareVersionNoDefault,
+		softwareVersionName,
+		softwareVersionValue,
```
kacf, line 311: What about `--software-filesystem`? This is supposed to be available for `module-image` as well.

artifact/metadata.go
```diff
@@ -295,9 +295,48 @@ func (a *ArtifactDepends) UnmarshalJSON(b []byte) error {
...
+	Name  string `json:"name"`
+	Value string `json:"value"`
+}
+
 type ArtifactProvides struct {
-	ArtifactName  string `json:"artifact_name"`
-	ArtifactGroup string `json:"artifact_group,omitempty"`
+	ArtifactName    string           `json:"artifact_name"`
+	ArtifactGroup   string           `json:"artifact_group,omitempty"`
+	SoftwareVersion *SoftwareVersion `json:"version,omitempty"`
```
kacf, line 306: I realize that is was not specified clearly enough in the task: We have two different `artifact_provides`, one in the global header, and one which is per payload. It is the payload header we want to change here, because it is a freeform structure, unlike this one. It's not super obvious in the documentation, but you can look at the difference in [the global one](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#artifact_provides) and [the payload one](https://github.com/mendersoftware/mender-artifact/blob/master/Documentation/artifact-format-v3.md#artifact_provides-1), which is inside `type-info` per payload.

Sorry, I missed this when you asked earlier in the chat. This will probably substantially change your PR, so I'm going to stop here for now.

kacf (changes requested):
A couple of global consideration:

* No changelogs! Please make sure this change is documented.
* Please target this to the `feature-partial_updates` branch, not master.

cli/mender-artifact/main.go
tranchitella: Added

artifact/metadata.go
tranchitella: I reworked my PR to write the software version attribute in the type-info.

tranchitella:
@kacf can you please review the PR before I fix the failing unit tests? They are caused by the extra provides.

kacf:
You can just update the tests to expect the new field. The new output looks correct.

tranchitella:
Manually launched pipeline with the following PRs:
* https://github.com/mendersoftware/integration/pull/1016
* https://github.com/mendersoftware/meta-mender/pull/1093

tranchitella:
New manual pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/193938184

tranchitella:
Depends on https://github.com/mendersoftware/meta-mender/pull/1096

tranchitella:
@kacf ready for another round of review, including Jira ticket MEN-3482

cli/mender-artifact/main.go
```diff
@@ -114,6 +121,24 @@ func getCliContext() *cli.App {
...
 	}
 
+	// Common Software Version flags
+	softwareVersionNoDefault := cli.BoolFlag{
+		Name:  noDefaultSoftwareVersionFlag,
+		Usage: "Disable the software version field for compatibility with old clients",
+	}
+	softwareVersionName := cli.StringFlag{
+		Name:  softwareNameFlag,
+		Usage: "Name of the key to store the software version: rootfs-image.NAME.version, instead of rootfs-image.version",
```
kacf, line 131: This help string is incorrect for `write module-image`. The default there is `rootfs-image.PAYLOAD_TYPE.version`.

cli/mender-artifact/write.go
```diff
@@ -333,6 +334,41 @@ func makeTypeInfo(ctx *cli.Context) (*artifact.TypeInfoV3, *artifact.TypeInfoV3,
...
+	softwareName := ctx.String(softwareNameFlag)
+	softwareVersion := ctx.String(softwareVersionFlag)
+	noDefaultSoftwareVersion := ctx.Bool(noDefaultSoftwareVersionFlag)
+	if softwareVersionMapping := getSoftwareVersion(artifactName, softwareFilesystem, softwareName, softwareVersion, noDefaultSoftwareVersion); len(softwareVersionMapping) > 0 {
+		if typeInfoProvides == nil {
+			typeInfoProvides = make(map[string]string)
+		}
+		for key, value := range softwareVersionMapping {
+			typeInfoProvides[key] = value
+		}
```
kacf: Since this is not clear from the function signature, I think there should be a comment on this function: "may modify the original map, or return a new one".

cli/mender-artifact/write.go
```diff
@@ -363,12 +364,34 @@ func applySoftwareVersionToTypeInfoProvides(ctx *cli.Context, typeInfoProvides a
...
+			}
 		}
 	}
 	return typeInfoProvides
 }
 
+func softwareVersionOverridesProvides(ctx *cli.Context, key string) bool {
+	cmdLine := strings.Join(os.Args, " ")
+
+	var providesVersion string = `(-p|--provides)(\s+|=)` + regexp.QuoteMeta(key) + ":"
```
kacf, line 383: Darn, I was hoping there would be a more straightforward way in urfave/cli.

cli/mender-artifact/write.go
```diff
@@ -363,12 +364,34 @@ func applySoftwareVersionToTypeInfoProvides(ctx *cli.Context, typeInfoProvides a
...
+	reProvidesVersion := regexp.MustCompile(providesVersion)
+	providesIndexes := reProvidesVersion.FindStringIndex(cmdLine)
```
kacf, line 385: Should use `FindAllStringIndex` and return the last one, otherwise this won't work: `-p X --software-version Y -p Z`. This may seem overkill, but we expect many of the cases where these flags are used to be automated, so there may be a lot of redundant flags (imagine all the redundant flags on a gcc line after running through some Makefile logic).

cli/mender-artifact/write.go
```diff
@@ -363,12 +364,34 @@ func applySoftwareVersionToTypeInfoProvides(ctx *cli.Context, typeInfoProvides a
...
+
+	var softareVersion string = "--software-(name|version|filesystem)"
+	reSoftwareVersion := regexp.MustCompile(softareVersion)
+	softwareIndexes := reSoftwareVersion.FindStringIndex(cmdLine)
```
kacf, line 389: Same here, `FindAllStringIndex`.

cli/mender-artifact/artifacts.go
```diff
@@ -330,13 +330,13 @@ func repack(comp artifact.Compressor, ua *unpackedArtifact, to io.Writer, key []
 		aWriter = awriter.NewWriterSigned(to, comp, artifact.NewSigner(key))
 	}
 
-	// for rootfs-images: Update rootfs_image_checksum provide if there is one.
-	_, hasChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs_image_checksum"]
+	// for rootfs-images: Update rootfs-image.checksum provide if there is one.
+	_, hasChecksumProvide := ua.writeArgs.TypeInfoV3.ArtifactProvides["rootfs-image.checksum"]
```
kacf, line 334: This is just modifying existing artifacts, so we should probably check for both here, and pass the correct legacy flag below.

cli/mender-artifact/write.go
```diff
@@ -333,6 +334,41 @@ func makeTypeInfo(ctx *cli.Context) (*artifact.TypeInfoV3, *artifact.TypeInfoV3,
...
 
+func getSoftwareVersion(artifactName, softwareFilesystem, softwareName, softwareVersion string, noDefaultSoftwareVersion bool) map[string]string {
+	result := map[string]string{}
+	softwareVersionName := "rootfs-image"
+	if softwareFilesystem != "" {
+		softwareVersionName = softwareFilesystem
+	}
+	if softwareName != "" {
+		softwareVersionName += fmt.Sprintf(".%s", softwareName)
+	}
```
kacf, line 351: For `write rootfs-image`, the default for `softwareVersionName` is `""`, but for `write module-image` it is the name of the payload type (`-T` parameter).

kacf (changes requested):

cli/mender-artifact/write.go
tranchitella: Done here: https://github.com/mendersoftware/mender-artifact/pull/291/commits/50ea75a6cd6e9e3d30311b24bfee258f32bc9209

cli/mender-artifact/artifacts.go
tranchitella: Done here: https://github.com/mendersoftware/mender-artifact/pull/291/commits/d0c71255610606ded4c1d8d5bf176a040baee2a8

cli/mender-artifact/write.go
tranchitella: Done here: https://github.com/mendersoftware/mender-artifact/pull/291/commits/50ea75a6cd6e9e3d30311b24bfee258f32bc9209

To be honest, I could not write a test because the `-p X --software-version Y -p Z` flags trigger a `argument specified more than once` error and X is applied instead of Z

cli/mender-artifact/write.go
tranchitella: See above

cli/mender-artifact/write.go
tranchitella: I couldn't find a way, the position of the flags is not exposed.

cli/mender-artifact/write.go
tranchitella: I updated the function to NOT modify the mapping passed in as argument, and added a comment.

cli/mender-artifact/main.go
tranchitella: Oh, sorry, I missed it.

kacf (approved):
Thanks!

cli/mender-artifact/write.go
kacf: That's weird, that's [being used in the tests](https://github.com/mendersoftware/mender-artifact/blob/master/cli/mender-artifact/dump_test.go#L165-L168).

tranchitella: @kacf yes, but there you are using different keys, aren't you?

kacf: Ah, good point. I think perhaps we should change this then. Being able to override the flags becomes more important with this new feature. Can you remove that message and change it to use the latest argument instead?

tranchitella: Done here: https://github.com/mendersoftware/mender-artifact/pull/291/commits/98ef1e96df8359736b09a99d68cb7ab4c677d31e

tranchitella:
New manual pipelline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/195271180

PR #294, opened 2020-09-24
MEN-3967: Fix segfault on mender-artifact dump for v2 Artifacts

lluiscampos:
Check the Artifact v3 specific fields before accessing them, this fixes
a segfault when running `mender-artifact dump --print-cmdline` on a v2
Artifact.

Changelog: Title

### DIFF
--- a/cli/mender-artifact/dump.go
+++ b/cli/mender-artifact/dump.go
@@ -183,19 +183,25 @@ func printCmdline(ar *areader.Reader, args []string) {
 	// this can recreate either type.
 	fmt.Printf("write module-image")
 
-	artProvs := ar.GetArtifactProvides()
-	fmt.Printf(" --artifact-name %s", artProvs.ArtifactName)
-	if len(artProvs.ArtifactGroup) > 0 {
-		fmt.Printf(" --provides-group %s", artProvs.ArtifactGroup)
-	}
+	if ar.GetInfo().Version == 3 {
+		artProvs := ar.GetArtifactProvides()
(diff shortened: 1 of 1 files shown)


oleorhagen (approved):
Another option could be to return an empty struct mby?

But this is fine.

cli/mender-artifact/dump.go
```diff
@@ -184,18 +184,22 @@ func printCmdline(ar *areader.Reader, args []string) {
 	fmt.Printf("write module-image")
 
 	artProvs := ar.GetArtifactProvides()
-	fmt.Printf(" --artifact-name %s", artProvs.ArtifactName)
-	if len(artProvs.ArtifactGroup) > 0 {
-		fmt.Printf(" --provides-group %s", artProvs.ArtifactGroup)
+	if artProvs != nil {
+		fmt.Printf(" --artifact-name %s", artProvs.ArtifactName)
```
kacf, line 188: I think there is a subtle error here. `ArtifactName` was not stored in provides in v2, it was stored in its own field. You might want to fall back to it.

The same goes for `device-type` below.

V2 is not important to keep around, so it's also ok not to bother with it, but if so, an error is a better outcome here, since the dump won't be correct.

lluiscampos: @kacf Thanks; good catch!

I decided for the former, correctly print `--artifact-name` and `--device-type` on v2 Artifacts

kacf (approved):

PR #248, opened 2019-12-23
added very simple execution tests for mac (& linux) 

mzedel:
this should just ensure the very basic functionality of the to-be-published binary
Changelog: None

Signed-off-by: Manuel Zedel <manuel.zedel@northern.tech>

PR #295, opened 2020-09-24
Add build:mac stage to the CI pipeline

alfrunes:
Trivial build step - only runs make inside the repository.
Already tested this here: https://gitlab.com/Northern.tech/Mender/mender-artifact/-/pipelines/194156353

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,6 @@
 stages:
-  - test
   - build
-  - smoketests
+  - test
   - publish
 
 include:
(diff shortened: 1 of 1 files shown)


lluiscampos (approved):
Awesome, "it works" :tada: 

We should review also the configuration in the machine, can @mzedel have a look at it?

mzedel (commented):
since we so far [publish](https://github.com/mendersoftware/mender-artifact/blob/master/.gitlab-ci.yml#L125-L129) the mac build from the cross platform build job we should either adjust the publishing as well or reuse the other mac build IMHO

alfrunes:
@mzedel @lluiscampos; In response to #248 I just merged the trivial build step into the smoke test.

> We should review also the configuration in the machine, can @mzedel have a look at it?

The configuration is as trivial as it gets for a basic `shell` executor, I only tweaked the concurrency level to 1 job at the time.

mzedel:
@alfrunes with the most recent changes we again don't test the binary we publish though... which is why I made the smoketests depend on the cross compilation build... `mender-artifact` is not exactly my backyard 😄 so feel free to decide what you want to do with it...

.gitlab-ci.yml
```diff
@@ -160,8 +157,9 @@ publish:tests:
 publish:s3:
   stage: publish
   image: debian:buster
-  dependencies:
-    - build:make
+  needs:
```
lluiscampos, line 160: Nice :+1:

alfrunes:
Thanks @mzedel, I didn't think of that.
I again redid the pipeline a little bit and added a `make build` command to the smoketests but reuse the published artifacts to actually do the smoketests. 

(Sorry about all the force pushes, I'm just unintentionally setting a new highscore in typos)

mzedel (approved):
👍

PR #303, opened 2020-10-05
Add the ability to choose a string separator for dump print-cmdline

oleorhagen:
This adds a flag 'separator', which can be set, in addition to print-cmdline,
which picks the separator to use for the output from the commmand.

The default is <space>.

Changelog: None

### DIFF
--- a/cli/mender-artifact/dump.go
+++ b/cli/mender-artifact/dump.go
@@ -15,6 +15,7 @@
 package main
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -91,7 +92,9 @@ func DumpCommand(c *cli.Context) error {
 	}
 
 	if c.Bool("print-cmdline") {
-		printCmdline(ar, dumpArgs)
+		printCmdline(ar, dumpArgs, false)
+	} else if c.Bool("print0-cmdline") {
+		printCmdline(ar, dumpArgs, true)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


cli/mender-artifact/main.go
```diff
@@ -543,6 +543,12 @@ func getCliContext() *cli.App {
 			Name:  "print-cmdline",
 			Usage: "Print the command line that can recreate the same Artifact with the components being dumped. If all the components are being dumped, a nearly identical Artifact can be created. Note that timestamps will cause the checksum of the Artifact to be different, and signatures can not be recreated this way. The command line will only use long option names.",
 		},
+		cli.StringFlag{
+			Name: "separator",
+			Usage: "Choose the separator to replace whitespace with for the 'print-cmdline' flag (default: <space>)",
```
kacf, line 548: Unfortunately being able to choose the separator won't work they way I had intended, because you cannot specify a nul character. What I had in mind was how the `-print0` operator works in [`find`](https://linux.die.net/man/1/find). The reason nul is particularly handy is that `xargs` supports the `-0` character which reads arguments terminated by nul characters, which would be very useful for shell handling. What about simply `print0-cmdline`, to mirror that?

cli/mender-artifact/dump.go
```diff
@@ -178,67 +179,69 @@ func dumpMetaData(metaDataDir string, dumpArgs *[]string, handlers map[int]handl
...
-			fmt.Printf(" --%s '%s'", clearsProvidesFlag, value)
+			fmt.Fprintf(buf, " --%s '%s'", clearsProvidesFlag, value)
 		}
 	}
 
 	if len(args) > 0 {
-		fmt.Printf(" %s\n", strings.Join(args, " "))
+		fmt.Fprintf(buf, " %s\n", strings.Join(args, " "))
 	}
+	fmt.Printf(strings.ReplaceAll(buf.String(), " ", separator))
```
kacf, line 244: This is not future proof, it will replace all spaces, including ones that are intended in a single argument. Granted, it's not what we're trying to tackle here, but I think it's better to get this right from the start. The change should be simple though, just print the separator at the start of every argument instead of the space.

kacf (changes requested):

cli/mender-artifact/main.go
oleorhagen: Alright, I misunderstood this a little

cli/mender-artifact/dump.go
```diff
@@ -178,67 +181,100 @@ func dumpMetaData(metaDataDir string, dumpArgs *[]string, handlers map[int]handl
...
+	var buf WriteStringer
+	if zeroSeparator {
+		buf = New()
+	} else {
+		buf = bytes.NewBuffer(nil)
+	}
 	// Even if it is a rootfs payload, we use the module-image writer, since
 	// this can recreate either type.
-	fmt.Printf("write module-image")
+	fmt.Fprintf(buf, "write module-image")
```
kacf, line 219: This also will not work, the separator has to be used *instead* of spaces, otherwise the spaces will become part of the argument. I was thinking more along the lines of:

```
sep := '\0'
fmt.Fprintf(buf, "write%cmodule-image", sep)
fmt.Fprintf(buf, "%c--artifact-name%c%s", sep, sep, artProvs.ArtifactName)
```

And so on.

cli/mender-artifact/main.go
```diff
@@ -543,6 +543,10 @@ func getCliContext() *cli.App {
 			Name:  "print-cmdline",
 			Usage: "Print the command line that can recreate the same Artifact with the components being dumped. If all the components are being dumped, a nearly identical Artifact can be created. Note that timestamps will cause the checksum of the Artifact to be different, and signatures can not be recreated this way. The command line will only use long option names.",
 		},
+		cli.BoolFlag{
+			Name:  "print0-cmdline",
+			Usage: "Same as 'print-cmdline', except that the arguments are separated by a null character (0x00).",
+		},
```
kacf, line 549: I guess it could also use a test.

cli/mender-artifact/dump.go
```diff
@@ -91,7 +92,9 @@ func DumpCommand(c *cli.Context) error {
 	}
 
 	if c.Bool("print-cmdline") {
-		printCmdline(ar, dumpArgs)
+		printCmdline(ar, dumpArgs, false)
+	} else if c.Bool("print0-cmdline") {
+		printCmdline(ar, dumpArgs, true)
```
kacf, line 97: Perhaps we should error if both conflicting arguments are given?

kacf (changes requested):

kacf:
I can fix this and resubmit, I'm back on this task again.

PR #305, opened 2020-10-13
MEN-3987 use sudo for snapshots if present.

merlin-northern:
Changes to sudo usage for snapshots; use sudo only if:

* sudo is present and works
* user is not root

ChangeLog:title

### DIFF
--- a/cli/mender-artifact/write.go
+++ b/cli/mender-artifact/write.go
@@ -520,8 +520,9 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 	// First echo to stdout such that we know when ssh connection is
 	// established (password prompt is written to /dev/tty directly,
 	// and hence impossible to detect).
-	args = append(args, "sudo", "-S", "/bin/sh", "-c",
-		`'echo "`+sshInitMagic+`" && mender snapshot dump | cat'`)
+	// When user id is 0 do not bother with sudo.
+	args = append(args, "/bin/sh", "-c",
+		`'[ $(id -u) -eq 0 ] || sudo_cmd="sudo -S"; $sudo_cmd /bin/sh -c "echo `+sshInitMagic+`; mender snapshot dump" | cat'`)
 
 	cmd := exec.Command("ssh", args...)
 


cli/mender-artifact/write.go
```diff
@@ -517,10 +528,28 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
...
 	args = append(args, userAtHost)
+	useSudo, _ := hasCommand(append(args, "sudo", "--help"))
+
+	_, out := hasCommand(append(args, "id", "-u"))
+	if string(out) == "0" {
+		useSudo = false
+		Log.Debug("effective user id on host is 0; not using sudo.")
+	}
+
+	if strings.HasPrefix(userAtHost, "root@") {
```
alfrunes, line 539: This should be an `else if` so as not to make the log entries confusing.

alfrunes (commented):
I thought of just by modifying the command to something along the lines of:
```
[ $UID -eq 0 ] && /bin/sh -c 'echo "<sshMagic>" && mender snapshot dump | cat' || sudo -S 'echo "<sshMagic>"' && sudo mender snapshot dump | cat'
```
But I'm fine with this solution as well, I just have one small comment.

cli/mender-artifact/write.go
```diff
@@ -486,6 +486,17 @@ func extractKeyValues(params []string) (*map[string]string, error) {
 	return keyValues, nil
 }
 
+func hasCommand(args []string) (bool, []byte) {
+	output, err := exec.Command("ssh", args...).Output()
+	if err == nil {
+		Log.Debugf("ssh %v: succesful; command present.", args)
+		return true, output
+	} else {
+		Log.Debugf("ssh %v: not succesful; assuming command not present.", args)
```
kacf, line 495: Minor: 2 x `s/succesful/successful/`

cli/mender-artifact/write.go
merlin-northern: sure. thanks.

alfrunes (approved):

cli/mender-artifact/write.go
```diff
@@ -517,10 +528,26 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 		args = append(args, "-p", port)
 	}
 	args = append(args, userAtHost)
+	useSudo, _ := hasCommand(append(args, "sudo", "--help"))
+
+	_, out := hasCommand(append(args, "id", "-u"))
```
kacf, line 533: A problem with this approach is that you have to type your ssh password three times. Can you work this into one call instead? For example something like:

```
args = append(args, "/bin/sh", "-c",
		`'if [ $(id -u) = 0 ]; then SUDO_CMD=sudo; else SUDO_CMD=; fi; $SUDO_CMD echo "`+sshInitMagic+`" && $SUDO_CMD mender snapshot dump | cat'`)
```

Untested and *probably* doesn't work exactly as I put it, but you get what I mean!

I think it is not important to handle the case where `id != 0` and `sudo` doesn't exist, because in this case the operation is guaranteed not to work anyway.

kacf (changes requested):

cli/mender-artifact/write.go
merlin-northern: you definitely broke all the fun here. I will try though.

kacf: Hehe, I know. But the User Is Always Right™! :slightly_smiling_face:

merlin-northern: hehe. right! I was _that_ close to actually ask for the password and play around with it, but in the end I followed [your advice](https://github.com/mendersoftware/mender-artifact/pull/305).

merlin-northern:
pipeline: https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/202444007

cli/mender-artifact/write.go
```diff
@@ -520,8 +520,8 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 	// First echo to stdout such that we know when ssh connection is
 	// established (password prompt is written to /dev/tty directly,
 	// and hence impossible to detect).
-	args = append(args, "sudo", "-S", "/bin/sh", "-c",
-		`'echo "`+sshInitMagic+`" && mender snapshot dump | cat'`)
+	args = append(args, "/bin/sh", "-c",
+		`'echo "`+sshInitMagic+`" && [ $(id -u) -eq 0 ] || sudo_cmd="sudo -S"; $sudo_cmd mender snapshot dump | cat'`)
```
kacf, line 524: @alfrunes can perhaps verify, but I think `sshInitMagic` needs to come after `sudo` because the client uses that string to identify the start of the payload data. If `sudo` comes after it, then there's a chance it will print `Password:`, which will then become part of your rootfs file instead of being displayed to the user.

merlin-northern: riight! got it! thanks

cli/mender-artifact/write.go
```diff
@@ -520,8 +520,9 @@ func getDeviceSnapshot(c *cli.Context) (string, error) {
 	// First echo to stdout such that we know when ssh connection is
 	// established (password prompt is written to /dev/tty directly,
 	// and hence impossible to detect).
-	args = append(args, "sudo", "-S", "/bin/sh", "-c",
-		`'echo "`+sshInitMagic+`" && mender snapshot dump | cat'`)
+	// When user id is 0 do not bother with sudo.
+	args = append(args, "/bin/sh", "-c",
+		`'[ $(id -u) -eq 0 ] || sudo_cmd="sudo -S"; $sudo_cmd /bin/sh -c "echo `+sshInitMagic+`; mender snapshot dump" | cat'`)
```
kacf, line 525: Small change, but it's finding the *right* change which is tricky! :+1:

kacf (approved):

cli/mender-artifact/write.go
merlin-northern: thanks Kristian!

PR #307, opened 2020-10-19
Fix: gocyclo refactoring and switch to go mod

tranchitella:
Changelog: none

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


tranchitella:
Stopping the QA pipeline, as it doesn't add any value in this PR.

lluiscampos (approved):

PR #308, opened 2020-10-19
Debug option to mender-artifact cli

merlin-northern:
Support for --debug:

```bash
#./mender-artifact write rootfs-image -f ssh://xxxx -n pi -o snapshot-release.1.0.mender -t rpi4 --debug
debug: creating artifact [snapshot-release.1.0.mender], version: 3
................................   0% 1024 KiB
```

ChangeLog:none

### DIFF
--- a/artifact/checksum.go
+++ b/artifact/checksum.go
@@ -16,7 +16,6 @@ package artifact
 
 import (
 	"bytes"
-	"github.com/minio/sha256-simd"
 	"encoding/hex"
 	"fmt"
 	"hash"
@@ -25,6 +24,7 @@ import (
 	"strings"
 	"syscall"
 
+	"github.com/minio/sha256-simd"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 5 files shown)


merlin-northern:
@oleorhagen I would like to add --debug to all the commands, and then migrate maybe to go-lib-micro log library. if you like that approach I will continue.

oleorhagen:
> @oleorhagen I would like to add --debug to all the commands, and then migrate maybe to go-lib-micro log library. if you like that approach I will continue.

I like this a lot! Logging in MA is long overdue. But can we break this up in two?

So maybe we can do logging first?

If so, why `go-lib-micro` instead of say, `logrus`, like we use in the client?

lluiscampos:
> If so, why `go-lib-micro` instead of say, `logrus`, like we use in the client?

I randomly caught this! Please let's stick to `logrus`. Otherwise we would need to package `go-lib-micro` for Debian repositories...

PR #324, opened 2021-01-18
MEN-4362: run fsck on fs image created via SSH snapshot

lluiscampos:
This can potentially fix a broken Artifact, and is harmless otherwise.

### DIFF
--- a/cli/mender-artifact/write.go
+++ b/cli/mender-artifact/write.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


cli/mender-artifact/write.go
```diff
@@ -119,6 +119,24 @@ func writeRootfs(c *cli.Context) error {
...
+		case ext:
+			err = runFsck(rootfsFilename, "ext4")
+		case unsupported:
+			err = errors.New("writeRootfs: unsupported filesystem")
+
+		}
+		if err != nil {
+			return cli.NewExitError("runFsck error: "+err.Error(), errArtifactCreate)
+		}
 	}
```
lluiscampos, line 140: I originally considered running `fsck` for all images at `writeRootfs`.

On the one hand many tests would require fixing, but on the other hand I realized it won't be a good idea anyway as it will be changing filesystem checksums. Hence my final conclusion on only "fixing" images generated via SSH.

lluiscampos:
I originally considered running `fsck` for all images at `writeRootfs`.

On the one hand many tests would require fixing, but on the other hand I realized it won't be a good idea anyway as it will be changing filesystem checksums. Hence my final conclusion on only attempting to fix images generated via SSH.

lluiscampos:
The drop in coverage is due to SSH snapshot feature being tested only in acceptance tests. See missing coverage from master [here](https://coveralls.io/builds/36390589/source?filename=cli%2Fmender-artifact%2Fwrite.go#L116).

kacf (approved):

PR #325, opened 2021-01-18
3.4.x: MEN-4362: run fsck on fs image created via SSH snapshot

lluiscampos:
Cherry-picking #324

### DIFF
--- a/cli/mender-artifact/write.go
+++ b/cli/mender-artifact/write.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


lluiscampos:
Both for `2.5.x` and `2.4.x`, I get the error:
```
Command 'mender-binary-delta-generator -n v2.0-deltafrom-v1.0 /tmp/meta-mender-acceptance.gx4dzdfb/artifact_from.mender /tmp/meta-mender-acceptance.gx4dzdfb/artifact_to.mender -o /tmp/meta-mender-acceptance.gx4dzdfb/artifact_delta.mender' returned non-zero exit status 1.
```

@kacf have you been able to run `TestDeltaUpdateModule::test_perform_update` test case on your recent release builds?

kacf:
Yes, this is why I did https://github.com/mendersoftware/mender-qa/pull/476. You need to set `MENDER_BINARY_DELTA_VERSION = 1.1.1` because 1.2.0 requires mender-artifact 3.5.0 or later.

It's just now dawning on me that this will break every automatically triggered job for the release branches... Any suggestions? We could hardcode in the trigger mechanism to use mender-binary-delta 1.1.1 for integration versions 2.5.x and older, but it seems tedious.

lluiscampos:
> Yes, this is why I did [mendersoftware/mender-qa#476](https://github.com/mendersoftware/mender-qa/pull/476). You need to set `MENDER_BINARY_DELTA_VERSION = 1.1.1` because 1.2.0 requires mender-artifact 3.5.0 or later.
> 

Thanks. Rerunning build (only for the failing platforms) at:
* https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/243659348
* https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/243659493

> It's just now dawning on me that this will break every automatically triggered job for the release branches... Any suggestions? We could hardcode in the trigger mechanism to use mender-binary-delta 1.1.1 for integration versions 2.5.x and older, but it seems tedious.

Yes, this dependency is very tricky. Hard coding the version for integration version <= 2.5 might not be that bad, as it is not something that we need to update in the future.

PR #328, opened 2021-02-05
Logo link fix

merlin-northern:
ChangeLog:none

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ The artifacts library might also be useful for other updaters or
 purposes. We are always happy to see other uses of it!
 
 
-![Mender logo](https://mender.io/user/pages/04.resources/logos/logoS.png)
+![Mender logo](https://hosted.mender.io/ui/assets/img/loginlogo.png)
 
 
 ## Getting started


estenberg:
My browser does not render https://hosted.mender.io/ui/assets/img/loginlogo.png

Maybe we should find a better one? Likely in mender.io somewhere.

Also, unfortunately, this will probably apply to *all* the repositories.. so maybe a redirect would be easier/better?

merlin-northern:
> My browser does not render https://hosted.mender.io/ui/assets/img/loginlogo.png
> 
> Maybe we should find a better one? Likely in mender.io somewhere.
> 
> Also, unfortunately, this will probably apply to _all_ the repositories.. so maybe a redirect would be easier/better?

wow, something is not right, I cant find the repo I took that from, also many other repos have the logo broken, as you pointed out. I have no idea, let me check.

oleorhagen:
@merlin close?

merlin-northern:
> @Merlin close?

the logo is still broken. I wonder when I will have time to get back to it.

PR #332, opened 2021-03-03
MEN-4502: Do not change the underlying Artifact unnecessarily

oleorhagen:
Fixes the issue where modifying an artifact by any of the utility commands would change the compression to `gzip`, as the Artifact was always repacked with the standard compression. This would also destroy signed Artifacts using commands such as `cat` which should have no effect on the signature.

This led to the discovery along the way, that most commands did indeed not respect the `compression` flag - which claimed to be global in scope. Therefore, this removes the global claim from the helptext, and instead adds a note about the `compression` flag's functionality, which after all, if you read the note, was limited to only a few commands in scope.

I guess most people never used the `compression` flag with most of the utility commands (as it would not have worked). Therefore I feel these changes are safe. 

The warning added tells the user to use  the `<modify>` command to change the `compression` type of an image/Artifact. Which I feel is, and should have been, the right approach all along.

### DIFF
--- a/cli/mender-artifact/artifacts.go
+++ b/cli/mender-artifact/artifacts.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
--- a/cli/mender-artifact/copy.go
+++ b/cli/mender-artifact/copy.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -22,17 +22,13 @@ import (
 	"path/filepath"
 	"regexp"
 
-	"github.com/mendersoftware/mender-artifact/artifact"
 	"github.com/urfave/cli"
 )
 
 var isimg = regexp.MustCompile(`\.(mender|sdimg|uefiimg|img)`)
 
 func Cat(c *cli.Context) (err error) {
-	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
-	if err != nil {
-		return cli.NewExitError("compressor '"+c.GlobalString("compression")+"' is not supported: "+err.Error(), 1)
-	}
+
 	if c.NArg() != 1 {
 		return cli.NewExitError(fmt.Sprintf("Got %d arguments, wants one", c.NArg()), 1)
 	}
@@ -45,7 +41,7 @@ func Cat(c *cli.Context) (err error) {
 		return cli.NewExitError("Unable to load key: "+err.Error(), 1)
 	}
 
-	r, err := virtualImage.OpenFile(comp, privateKey, c.Args().First())
+	r, err := virtualImage.OpenFile(privateKey, c.Args().First())
 	defer func() {
 		if r == nil {
 			return
@@ -66,9 +62,9 @@ func Cat(c *cli.Context) (err error) {
 }
 
 func Copy(c *cli.Context) (err error) {
-	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
-	if err != nil {
-		return cli.NewExitError("compressor '"+c.GlobalString("compression")+"' is not supported: "+err.Error(), 1)
+	if c.String("compression") != "" {
+		fmt.Fprintf(os.Stderr, "Warning: The compression flag is not respected for the copy command.\n"+
+			"If you wish to change the compression type, use the <modify> command.")
 	}
 
 	privateKey, err := getKey(c.String("key"))
@@ -95,7 +91,7 @@ func Copy(c *cli.Context) (err error) {
 		if err != nil {
 			return cli.NewExitError(err, 1)
 		}
-		vfile, err = virtualImage.OpenFile(comp, privateKey, c.Args().Get(1))
+		vfile, err = virtualImage.OpenFile(privateKey, c.Args().Get(1))
 		defer wclose(vfile)
 		if err != nil {
 			return cli.NewExitError(err, 1)
@@ -106,14 +102,14 @@ func Copy(c *cli.Context) (err error) {
 		return nil
 	case copyinstdin:
 		r = os.Stdin
-		vfile, err = virtualImage.OpenFile(comp, privateKey, c.Args().Get(1))
+		vfile, err = virtualImage.OpenFile(privateKey, c.Args().Get(1))
 		defer wclose(vfile)
 		if err != nil {
 			return cli.NewExitError(err, 1)
 		}
 		w = vfile
 	case copyout:
-		vfile, err = virtualImage.OpenFile(comp, privateKey, c.Args().First())
+		vfile, err = virtualImage.OpenFile(privateKey, c.Args().First())
 		defer wclose(vfile)
 		if err != nil {
 			return cli.NewExitError(err, 1)
@@ -140,10 +136,6 @@ func Copy(c *cli.Context) (err error) {
 // Install installs a file from the host filesystem or directory onto either
 // a mender artifact, or an sdimg.
 func Install(c *cli.Context) (err error) {
-	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
-	if err != nil {
-		return cli.NewExitError("compressor '"+c.GlobalString("compression")+"' is not supported: "+err.Error(), 1)
-	}
 
 	privateKey, err := getKey(c.String("key"))
 	if err != nil {
@@ -168,7 +160,7 @@ func Install(c *cli.Context) (err error) {
 		}
 		perm = os.FileMode(c.Int("mode"))
 		if directory {
-			vdir, err := virtualImage.OpenDir(comp, privateKey, c.Args().First())
+			vdir, err := virtualImage.OpenDir(privateKey, c.Args().First())
 			defer wclose(vdir)
 			if err != nil {
 				return cli.NewExitError(err, 1)
@@ -198,7 +190,7 @@ func Install(c *cli.Context) (err error) {
 			}
 		}()
 
-		vfile, err := virtualImage.OpenFile(comp, privateKey, c.Args().Get(1))
+		vfile, err := virtualImage.OpenFile(privateKey, c.Args().Get(1))
 		defer wclose(vfile)
 		if err != nil {
 			return cli.NewExitError(fmt.Sprintf("%v", err), 1)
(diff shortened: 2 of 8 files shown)


oleorhagen:
@mender-test-bot start pipeline please!

lluiscampos (commented):
I am sorry but you need to ELI5 this to me... I don't manage to follow the existing code to see why closing the files is a problem.

lluiscampos:
Thanks for the call, I understood it now.

As discussed: the problem is in `repackArtifact` and we have fixed `cat` and `copy` (out) but just avoiding the close and therefore `repackArtifact` -> However I suspect we have the same problem on `copy` (in), `install`, `delete`, and any other command that has to actually call `repackArtifact`

oleorhagen:
> Thanks for the call, I understood it now.
> 
> As discussed: the problem is in `repackArtifact` and we have fixed `cat` and `copy` (out) but just avoiding the close and therefore `repackArtifact` -> However I suspect we have the same problem on `copy` (in), `install`, `delete`, and any other command that has to actually call `repackArtifact`

Well, not the same problem per se. This is unrelated to the `cat` and `copy` and `modify` commands, and lies deeper in the parser. I've had a look at it, and it is not such an easy fix, since it touches a lot of other parts.

I say we merge this, independently, and then create a follow up for the re-packaging with the proper compression format, in the meantime, it is possible to force the correct compression using something like:
```
mender-artifact cp <from> <to> --compression=<correct-compression>
```
What do you say?

oleorhagen:
Actually on the trail of something now

oleorhagen:
@lluiscampos have a look at the latest 2 commits, and see if you think this is sane or not

lluiscampos:
> @lluiscampos have a look at the latest 2 commits, and see if you think this is sane or not

It makes much sense to me. Getting the compression parameter from the reader is the right thing to do :+1:  I'll have a second read on the code in a short while.

> ...
> What do you say?

Merging as it was would have worked too, you had a valid point.

oleorhagen:
Great, I will probably squash the two into one then, and add a `Changelog` to begin with.

oleorhagen:
Tests will also need a little polishing

cli/mender-artifact/copy.go
```diff
@@ -22,17 +22,12 @@ import (
...
 	"github.com/urfave/cli"
 )
 
 var isimg = regexp.MustCompile(`\.(mender|sdimg|uefiimg|img)`)
 
 func Cat(c *cli.Context) (err error) {
-	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
-	if err != nil {
-		return cli.NewExitError("compressor '"+c.GlobalString("compression")+"' is not supported: "+err.Error(), 1)
-	}
```
kacf, line 35: I think this block (and the other ones you changed) should be replaced with a warning instead then. Something like "`--compression flag ignored, using existing compression type`" or something like that.

cli/mender-artifact/artifacts.go
```diff
@@ -326,7 +326,7 @@ func reconstructArtifactWriteData(ua *unpackedArtifact) (*awriter.WriteArtifactA
 }
 
 func repack(comp artifact.Compressor, ua *unpackedArtifact, to io.Writer, key []byte) error {
-	aWriter := awriter.NewWriter(to, comp)
+	aWriter := awriter.NewWriter(to, ua.ar.Compressor())
```
kacf, line 329: Does this prevent repacking with a new compression type using `modify`? Because this is a quite important use case, if you're trying to deploy to a device and find that it's lacking lzma, for example.

cli/mender-artifact/copy.go
```diff
@@ -46,15 +46,6 @@ func Cat(c *cli.Context) (err error) {
...
 	r, err := virtualImage.OpenFile(comp, privateKey, c.Args().First())
-	defer func() {
-		if r == nil {
-			return
-		}
-		cerr := r.Close()
-		if err == nil {
-			err = cerr
-		}
-	}()
```
kacf, line 57: I haven't checked explicitly, but I suspect this will leave all temporary (and very big!) files on the filesystem, so you should check this.

If my suspicion is correct, then a better fix might be to leave this in, but instead have a flag in `r` which indicates whether you want repacking when the handle is closed or not.

cli/mender-artifact/copy.go
oleorhagen: Hmm, yeh, maybe. This one's a little annoying too though, as the flag always has a default `gzip` value, has it not? So the warning will be printed on every single invocation

cli/mender-artifact/artifacts.go
oleorhagen: I see, hmm, then we will need a special case for this

cli/mender-artifact/copy.go
oleorhagen: It will -_-
I was hoping for a quick-fix.
Some handlers already have a `dirty` flag set, when closing. I will extend this to all the other handlers too.

cli/mender-artifact/copy.go
kacf: Yeah, but maybe it can be changed to empty default instead? Which is taken into account when the compressor is created.

cli/mender-artifact/main.go
```diff
@@ -478,12 +478,6 @@ func getCliContext() *cli.App {
 		Action: Copy,
 	}
 
-	copy.Flags = []cli.Flag{
-		compressionFlag,
-		privateKeyFlag,
-	}
```
oleorhagen, line 484: Hmm, this looks a little too much to me now, looking at it.
I guess at least the `privateKeyFlag` should stay? And possibly the `compressionFlag` is needed too, in order to actually print the warning when it is set?

cli/mender-artifact/modify.go
```diff
@@ -46,7 +46,12 @@ func modifyArtifact(c *cli.Context) (err error) {
 		return cli.NewExitError("File ["+c.Args().First()+"] does not exist.", 1)
 	}
 
-	image, err := virtualImage.Open(comp, privateKey, c.Args().First())
+	var image VPImage
+	if c.String("compression") != "" {
+		image, err = virtualImage.Open(privateKey, c.Args().First(), comp)
+	} else {
+		image, err = virtualImage.Open(privateKey, c.Args().First())
+	}
```
oleorhagen, line 54: Not pretty, but it was the least invasive way I found around this issue

cli/mender-artifact/modify.go
```diff
@@ -62,6 +67,7 @@ func modifyArtifact(c *cli.Context) (err error) {
 		}
 	}()
 
+	image.dirtyImage()
```
oleorhagen, line 70: I decided not to be too surgical with `modify`, as the odds are every command will need the `Artifact` to be repacked.

cli/mender-artifact/partition.go
```diff
@@ -49,6 +49,7 @@ type VPImage interface {
 	io.Closer
 	Open(fpath string) (VPFile, error)
 	OpenDir(fpath string) (VPDir, error)
+	dirtyImage()
```
oleorhagen, line 52: Now only repack images when a command has called `dirtyImage()` as a part of it's execution. This means every command like `CopyTo`, `Write`, etc, all need to call this upon execution.

For this PR, I tried plumbing everything from the image wrappers: `imageAndFile` etc.

cli/mender-artifact/partition.go
```diff
@@ -130,6 +132,13 @@ func (v vImage) Open(comp artifact.Compressor, key []byte, imgname string) (VPIm
 			return nil, errors.Wrap(err, "can not process artifact")
 		}
 
+		var comp artifact.Compressor
+		if len(overrideCompressor) == 1 {
+			comp = overrideCompressor[0]
```
oleorhagen, line 137: Don't really think it's necessary to be too defensive here, we control everywhere this is called.

oleorhagen:
@kacf @lluiscampos I have taken another stab at this (it's been - as always when working with this repo _sigh_ - slightly painful do get anything done without creating a complete mess). 

Please have another look, and tell me what you think of the approach, I've tried to be as clear as I can, but my head's in a little too deep atm, and I'm struggling with getting perspective. Hopefully it will be easier for you guys to see some obvious mistakes.

I've left some comments to help guide you through the review.

Cheers
- Ole uno

cli/mender-artifact/main.go
oleorhagen: Well, for now, the `compression` flag does not apply when copying into the Artifact. If compression is set, a warning is issued though, that you should use `mender-artifact modify` to change the compression of your `Artifact`.

cli/mender-artifact/modify.go
oleorhagen: Nope, because that is the compressor is there acquired from the `global` scope variable for `compression`, which defaults to `gzip`.

kacf: :+1:

cli/mender-artifact/main.go
kacf: Ok, sounds good!

Discussed on Slack: Keep both flags, but warn on use of `compression`.

cli/mender-artifact/main.go
```diff
@@ -59,6 +59,38 @@ func main() {
 	}
 }
 
+// Copied from urfave/cli/template.go
```
lluiscampos, line 62: Maybe we could have updated [our fork](https://github.com/mendersoftware/cli/commit/2eb686fe967b14b3e204f0eff66427e1fdafcf6d) instead?

oleorhagen: Hmm, possibly, I'm not sure If I think it's a better solution though. Let's try and keep the forks as similar as possible

cli/mender-artifact/copy.go
```diff
@@ -66,9 +62,9 @@ func Cat(c *cli.Context) (err error) {
 }
 
 func Copy(c *cli.Context) (err error) {
-	comp, err := artifact.NewCompressorFromId(c.GlobalString("compression"))
-	if err != nil {
-		return cli.NewExitError("compressor '"+c.GlobalString("compression")+"' is not supported: "+err.Error(), 1)
+	if c.String("compression") != "" {
+		fmt.Fprintf(os.Stderr, "Warning: The compression flag is not respected for the copy command.\n"+
+			"If you whish to change the compression type, use the <modify> command.")
```
kacf, line 67: `s/whish/wish/`

cli/mender-artifact/main.go
kacf: Agree, this is a custom template, we should not modify the original I think.

kacf (approved):
LGTM, just one minor typo.

oleorhagen:
@mender-test-bot start pipeline :robot:

oleorhagen:
Only `test_bogus_proto` test failed in `OS`. 

Good to merge as is.

oleorhagen:
Cherry: 3.5.x: https://github.com/mendersoftware/mender-artifact/pull/334

PR #333, opened 2021-03-16
[Cherry] MEN-4502 3.4.x

oleorhagen:


oleorhagen:

kacf:
I would consider this a quite risky bug fix for a patch release, it's quite intrusive. Do we have a strong incentive to backport it?

oleorhagen:
> I would consider this a quite risky bug fix for a patch release, it's quite intrusive. Do we have a strong incentive to backport it?

Well, you know the bug as well as I do, and if you don't think it's worthwhile, then I'm sure it's not.

I initially thought it could be pretty confusing, and the bug is pretty unrelated to the other functionality of `mender-artifact`.

But your call

kacf:
I think my vote would be to omit it. It's not really a critical fix, I'd rather avoid the risk for this branch if we can.

oleorhagen:
Closed: Decided the port is not critical

PR #335, opened 2021-03-19
[copy_test] Amend tests to always call verifyTestFunc

lluiscampos:
This function is also used in most test case to clean-up temp files, so
remove the early return on test cases expecting error, so that we call
such function for all.

Introduced at a4bf948c.


lluiscampos:
I discovered this because of a warning [here](https://lintian.debian.org/sources/golang-github-mendersoftware-mender-artifact/3.4.0-1.html). That `input.txt` file is distributed today in `golang-github-mendersoftware-mender-artifact-dev` package.

oleorhagen (approved):
:+1:

PR #340, opened 2021-04-27
Cleanup: Delete the mender-artifact subdir from cli/mender-artifact

oleorhagen:
This looks huge, but really there isn't much to it:

* It removes the extra subdir ./cli/`mender-artifact`, which it seems only exists to have the name of the output binary be `mender-artifact`, since main lives in `./cli/mender-artifact`.
* It moves `main.go` to the top level directory
* Cleanups in the `Makefile` as a result of the above two
* Note: the binary will now be placed in the top-level directory by default, as opposed to `./cli/mender-artifact`. I'm not yet fully sure of the implications here.


oleorhagen:

lluiscampos (approved):
Excellent.

I reviewed the Makefile, the Dockerfile(s), etc and I don't see any side effects. All tests passed so that also give me confidence that the change is safe.

Thanks!

PR #344, opened 2021-07-05
[MEN-4832] Fix sending on closed signal channel

alfrunes:
There is a [potential race condition](https://golang.org/src/os/signal/signal.go#L199)
where signals sent right before or while signal.Stop is closed does not
reach the signal channel until after the function returns. In such cases
it can happen that the signal channel is closed before the signal has
been propagated to the handler. This commit instead of closing the
signal channel, uses a separate channel (context.WithCancel) to stop
listening for signals after calling signal.Stop.

changelog: title

### DIFF
--- a/cli/util/tty.go
+++ b/cli/util/tty.go
@@ -17,8 +17,8 @@
 package util
 
 import (
+	"context"
 	"os"
-	"os/signal"
 
 	"github.com/pkg/errors"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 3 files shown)


alfrunes:

lluiscampos (approved):
This is a bit beyond my technical understanding, but the use of `context` looks correct to me here.

alfrunes:
(I missed an import in the windows build)

alfrunes:
I don't think the test failures are related:
> Failed 16 times in master in the last 14 days

@lluiscampos I leave it to you to merge this PR.

PR #350, opened 2021-08-28
Improve install targets in Makefile

414n:
This aims to improve a couple of shortcomings in the Makefile related to file installation system-wide (or packaging-wide) that I noticed while creating a Slackware package for this program. These are:

 - improper PREFIX variable usage, i.e. using `$(PREFIX)/usr/local` will create a `/usr/local/usr/local` or `/usr/usr/local` hierarchy depending on whether PREFIX is set to `/usr` or `/usr/local`, which is probably not desirable;
 - inconsistent use of braces instead of round brackets.

I also added a couple of lines in the README.md file to mention these adjustable parameters, but do tell me if that falls in the "too much information in the README" category.

I tested these changes by simply pointing to a DESTDIR directory under `/tmp` and checking that the files are indeed put in the expected paths:

```sh
$ ls -l /tmp/test
ls: cannot access '/tmp/test': No such file or directory
$ make DESTDIR=/tmp/test PREFIX=/usr install-autocomplete-scripts
Installing Bash auto-complete script into /tmp/test/etc/bash_completion.d/
Installing zsh auto-complete script into /tmp/test/usr/share/zsh/site-functions/
$ tree /tmp/test/
/tmp/test/
├── etc
│   └── bash_completion.d
│       └── mender-artifact
└── usr
    └── share
        └── zsh
            └── site-functions
                └── _mender-artifact

6 directories, 2 files
```

If you're okay with these changes, I'd also like to add the capability to the `install` target to install the binary directly in `$(DESTDIR)$(PREFIX)/bin` instead of the default `$GOPATH/bin`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ PREFIX ?= /usr/local
 PKGS = $(shell go list ./... | grep -v vendor)
 SUBPKGS = $(shell go list ./... | sed '1d' | tr '\n' ',' | sed 's/,$$//1')
 PKGNAME = mender-artifact
+GOEXE := $(shell $(GO) env GOEXE)
+FINALEXE = $(PKGNAME)$(GOEXE)
 PKGFILES = $(shell find . \( -path ./vendor -o -path ./Godeps \) -prune \
 		-o -type f -name '*.go' -print)
 PKGFILES_notest = $(shell echo $(PKGFILES) | tr ' ' '\n' | grep -v _test.go)
@@ -37,8 +39,8 @@ ifneq ($(TAGS),)
 BUILDTAGS = -tags '$(TAGS)'
 endif
 
-build:
-	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
+build $(FINALEXE):
+	$(GO) build $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS) -o $(FINALEXE)
 
 PLATFORMS := darwin linux windows
 
@@ -68,8 +70,13 @@ build-natives-contained:
 	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact* /binary" && \
 	docker image rm $$image_id
 
-install:
+goinstall:
 	@$(GO) install $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
+
+install-all: install install-autocomplete-scripts
+
+install: $(FINALEXE)
+	install -Dm0755 $(FINALEXE) $(DESTDIR)$(PREFIX)/bin/$(FINALEXE)
 
 install-autocomplete-scripts:
 	@echo "Installing Bash auto-complete script into $(DESTDIR)/etc/bash_completion.d/"
@@ -84,6 +91,7 @@ clean:
 	$(GO) clean
 	rm -f mender-artifact-darwin mender-artifact-linux mender-artifact-windows.exe
 	rm -f coverage.txt coverage-tmp.txt
+	rm -f $(FINALEXE)
 
 get-tools:
 	set -e ; for t in $(TOOLS); do \
@@ -141,4 +149,4 @@ coverage:
 
 .PHONY: build clean get-tools test check \
 	cover htmlcover coverage tooldep install-autocomplete-scripts \
-	instrument-binary
+	instrument-binary install-all install
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ goinstall:
 
 install-all: install install-autocomplete-scripts
 
-install: $(FINALEXE)
+install: build
 	install -Dm0755 $(FINALEXE) $(DESTDIR)$(PREFIX)/bin/$(FINALEXE)
 
 install-autocomplete-scripts:
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,5 @@ RUN apt-get update && apt-get install -qyy liblzma-dev
 RUN mkdir -p /go/src/github.com/mendersoftware/mender-artifact
 WORKDIR /go/src/github.com/mendersoftware/mender-artifact
 ADD ./ .
-RUN make build
-RUN make install
+RUN make goinstall
 ENTRYPOINT [ "/go/bin/mender-artifact" ]


kacf (changes requested):
Code wise, this looks great, it's good to have these cleaned up!

One thing needs fixing in the commit. For legal/technical reasons, the signoff string needs to match the author string exactly.

kacf:
@lluiscampos: If we are ever in doubt in the other repositories, I believe this is how it should be done! :-)

lluiscampos:
@414n thanks for your contribution :+1: 

> @lluiscampos: If we are ever in doubt in the other repositories, I believe this is how it should be done! :-)

@kacf Indeed. This is the way yocto defines it too.

414n:
> Code wise, this looks great, it's good to have these cleaned up!
> 
> One thing needs fixing in the commit. For legal/technical reasons, the signoff string needs to match the author string exactly.

Thanks, I've tried a couple of times to fix the `Signed-off-by` but did not understand what else was failing :bowing_man: 

Will now look into streamlining a bit the `install` target too.

kacf (approved):
Thanks! Will run this through the full test suite now.

414n:
I refactored a bit the `install` target too.
While doing that, I did some enhancements to the docker related targets too.
Maybe I went a bit too far, so please tell me if you prefer to revert be1d4282.

Makefile
```diff
@@ -57,21 +73,17 @@ build-natives:
 	@env GOOS=windows GOARCH=$$arch CC=x86_64-w64-mingw32-gcc CXX=x86_64-w64-mingw32-g++ \
 		$(GO) build $(GO_LDFLAGS_WIN) $(BUILDV) -tags $(TAGS) nolzma -o $(PKGNAME)-windows.exe ;
 
-build-contained:
-	rm -f mender-artifact && \
-	image_id=$$(docker build -f Dockerfile . | awk '/Successfully built/{print $$NF;}') && \
-	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact /binary" && \
-	docker image rm $$image_id
+build-contained: $(DOCKER_BUILD_FLAG_BASE)
```
kacf, line 76: This doesn't take into account that the source might have changed. I guess it's fixable, but checking for all the source file will complicate it a bit. I guess the question is whether it's worth the hassle.

Makefile
```diff
@@ -57,21 +73,17 @@ build-natives:
...
-	image_id=$$(docker build -f Dockerfile.binaries . | awk '/Successfully built/{print $$NF;}') && \
-	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact* /binary" && \
-	docker image rm $$image_id
+build-natives-contained: $(DOCKER_BUILD_FLAG_NATIVES)
+	docker run --rm --entrypoint "/bin/sh" -u $$UID:$(shell id -g) -v $(shell pwd):/binary $(DOCKER_IMAGE_NAME_NATIVES) -c "cp /go/bin/$(PKGNAME)* /binary"
 
 goinstall:
 	@$(GO) install $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
+install-all: install install-autocomplete-scripts
```
kacf, line 85: This should be in a separate commit.

Makefile
```diff
@@ -67,9 +69,12 @@ build-natives-contained:
 	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact* /binary" && \
 	docker image rm $$image_id
 
-install:
+goinstall:
 	@$(GO) install $(GO_LDFLAGS) $(BUILDV) $(BUILDTAGS)
 
+install: $(FINALEXE)
+	install -Dm0755 $(FINALEXE) $(DESTDIR)$(PREFIX)/bin/$(FINALEXE)
```
kacf, line 88: We've had bad experience with this in the past, see [this commit](https://github.com/mendersoftware/mender/commit/b6e20f1f274e55fc09c9b0c09b5c8f9429ca2b05). `go install` caches the build artifacts, so I don't think this is very important for performance. Unless there's a strong reason I think it's better to keep this the same across the Mender repositories.

kacf (changes requested):
FYI, I merged the commits we already agreed on.

I'm a bit more sceptical of the new commits. Take a look at the comments and we can take it from there.

Makefile
414n: Yes, you're right, at the moment it only looks for changes in the related Dockerfile.
I'm still looking into making it aware of Go source file changes too but, as you said, it probably won't be easy as Go projects structure is not really Makefile-friendly. If you deem this as overcomplicated/overengineered, I can leave it as that by reverting the last commit.

Makefile
414n: I'm sorry, it should've went into 3733b77bcc8d01ed7e5be2c7f73e1131e1f84932.
Will fix it ASAP

Makefile
414n: I think there's a misunderstanding here. I changed the `install` target mainly to map it to the action of installing the final binary in a standard system-wide location (`$PREFIX/bin`, optionally prefixed with a `$DESTDIR` path useful when packaging), while before it was just called to build the final binary and put it under `$GOPATH/bin`.
Now, by looking at mendersoftware/mender@b6e20f1f2, I can see the concern regarding build tags change detection that can be mismanaged by only adding a target dependency on `FINALEXE`.
I can definitely look into it.

Makefile
kacf: Yes, given that the benefit is pretty low (Go already caches build artifacts), I think I would vote for omitting this to keep the complexity down.

Makefile
kacf: Yes, the Go command would not obey `PREFIX` and `DESTDIR`.

I think this can be solved by simply replacing the dependency on `$(FINALEXE)` with `build` instead, and then making sure that `build` does not depend on it either (to avoid short circuiting the build).

Makefile
414n: Force pushed the change just now.
Sorry it took so long.

Makefile
414n: Removed the `$(FINALEXE)` dependency in favor of the phony `build` target dependency: this should enforce a fresh build on every `install` invocation.

Makefile
kacf: This commit should be removed, as per the discussion.

There is definitely room for improvement here, but we can't submit something that doesn't take code changes into account.

kacf (changes requested):
The first and last commit are ready to go as-is. I suggest you just remove the "enhance docker targets", so we can get those two commits merged. If you still want to work on "enhance docker targets", it can be done separately.

414n:
Ok, got rid of the docker targets "enhancements" and kept only the other two commits, as suggested.

Let me know if anything needs further polishing.

kacf (approved):
Sorry for the delay. Yes, it looks good now! I'll run this through our test framework.

PR #355, opened 2021-09-22
MEN-5094: Empty the tar reader to ensure a full read

oleorhagen:
This ensures that the checksum calculated is correct.

### DIFF
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -26,10 +26,11 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/pkg/errors"
+
 	"github.com/mendersoftware/mender-artifact/artifact"
 	"github.com/mendersoftware/mender-artifact/handlers"
 	"github.com/mendersoftware/mender-artifact/utils"
-	"github.com/pkg/errors"
 )
 
 type SignatureVerifyFn func(message, sig []byte) error
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -160,6 +160,10 @@ func (ar *Reader) readHeader(headerSum []byte, comp artifact.Compressor) error {
 		return err
 	}
 
+	// Empty the remaining reader
+	// See (MEN-5094)
+	io.Copy(ioutil.Discard, r)
+
 	// Check if header checksum is correct.
 	if cr, ok := r.(*artifact.Checksum); ok {
 		if err = cr.Verify(); err != nil {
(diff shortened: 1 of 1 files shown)


oleorhagen:

areader/reader.go
```diff
@@ -160,6 +160,9 @@ func (ar *Reader) readHeader(headerSum []byte, comp artifact.Compressor) error {
 		return err
 	}
 
+	// Empty the remaining reader
+	io.Copy(ioutil.Discard, r)
```
kacf, line 164: Nice fix!

The fact that it needs to be called here is quite surprising though, it would be nice to have a ticket reference here for later, to understand why it is needed. Same for below.

kacf (approved):
This should have a changelog, since it fixes a bug reported by a user.

areader/reader.go
oleorhagen: So, I'm not really happy with this fix tbh. But the checksum reader and the verify function is sort of entangled too, so, yeh. this was the least intrusive fix I could find. But it does not fix this error for other parts using the checksum reader...

I don't know really..

Maybe I should dive deeper with this (but the reader code is messy :anger: )

grandfield: > What is the bytes left over in the reader?

The header.tar.gz is 514 bytes, with the last two bytes being 00 00. Sometimes the read was 514, other times the read was 512.

grandfield: > No test, I haven't been able to reproduce.

It is not reproducible in standalone, only via daemon downloading from AWS s3.

kacf: @grandfield: And can you reproduce it with this fix?

@oleorhagen: The fact that it's an elusive bug is why I thought a comment would be good. It at least reduces the chance that we will remove it by mistake. I think it will be hard to test this when it's not consistently reproducible.

oleorhagen: Added the comment and the changelog.

kacf (approved):

areader/reader.go
grandfield: > @grandfield: And can you reproduce it with this fix?
Works for me consistently now.

lluiscampos:
The [previous pipeline](https://gitlab.com/Northern.tech/Mender/mender-qa/-/pipelines/375013271) was successful, the last push only added a comment in the code. Merging.

PR #359, opened 2021-09-23
3.6.x: MEN-5094: Empty the header reader before calculating the checksum

lluiscampos:
Porting #355.


lluiscampos:
I am not running the full pipeline for this PR, as the upcoming release candidate will test it.

PR #360, opened 2021-09-27
MEN-5082: Restore SSH snapshot feature on Mac OS

lluiscampos:
While fixing MEN-4362, we introduced a dependency for the SSH snapshot
feature on `blkid`, in effect introducing a regression on Mac OS where
this tool is not present.

This commit amends 3457a15 converting it into a soft requirement: run
fsck only if supported, leaving otherwise the Artifact unaltered (which
the underlying risk that could be corrupted)

Changelog: Title

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -59,7 +59,13 @@ build:make:
     paths:
       - mender-artifact-*
   tags:
-    - docker
+    - mender-qa-slave
+  variables:
+    # Docker dind configuration for mender-qa-slave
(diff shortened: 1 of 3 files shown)


lluiscampos:
I am a bit unsure about the warning message, should we just silently skip?

kacf (approved):

mzedel (approved):
👍 just gave this a try and got `warning: Skipping running fsck on the Artifact: `blkid` binary not found on the system`, followed by the artifact writing steps

PR #364, opened 2021-11-11
Create a new signer using GCP's KMS

Michaelhobo:
This signs and verifies data using GCP's Key Management Service. This
allows developers to use mender-artifact without ever accessing the
private signing key.


Michaelhobo:
Sorry for the vendor/ noise.

Michaelhobo:
I realized that Google KMS returns a signature in a different format, not the one used by Mender. My guess is it's in the ASN1 format, based on this guide: https://cloud.google.com/kms/docs/create-validate-signatures#validate_ec_signature

I'll have to convert from ASN1 to Mender's serialization format for ECDSA signatures. While I fix that, comments on the rest are still welcome!

Michaelhobo:
@kacf I don't see a doc for how to contribute, and I don't know how to add a reviewer. Sorry for roping you in.

It seems like we are tied to go 1.14, and I'm using `crypto/ecdsa`'s `SignASN1` function in a test, which doesn't exist in go 1.14. Any chance we could rev the go version to 1.15 or later? 😄 

I also don't know how to fix the other CI failures (check-commits and check-license).

kacf:
No problem @Michaelhobo!

Since this library is used by the Mender client, we are tied to the Golang version of the oldest Yocto release we support. Currently, this is "warrior", which is using an even older Golang release: 1.12. Looping in @lluiscampos, in case he has an opinion on this.

I'm also a bit worried about the binary size from the vendor dependency, which is enormous. It is possible I'm wrong though, and it would be good to get some hard numbers on it. Could you compile the Mender client normally, then update the vendored mender-artifact to the one you just made, compile it again, and then compare the sizes?

Assuming the filesize has indeed increased substantially, my suggestion for dealing with both of these issues would be: The client only uses the "mender-artifact/areader" package, and packages used indirectly from there. But currently this includes the new dependency because of the indirect reference to "mender-artifact/artifact". My suggestion is that you put your new code into a package of its own, and instead of importing it directly in signer.go, use a signer-interface as a function argument. This can be passed in for example from "mender-artifact/cli", which isn't used by the client.

kacf:
As for the other errors, please read [our contribution guide](https://github.com/mendersoftware/mender/blob/master/CONTRIBUTING.md). It should explain what to do to fix the "check_commits" failure.

The license check is failing because we need to track all licenses in `LIC_FILES_CHKSUM.sha256`. You need to verify that all the licenses of the new vendored components are permissive licenses (IOW one of the types that already exist inside `LIC_FILES_CHKSUM.sha256`), and then add them to that file.

lluiscampos:
> Since this library is used by the Mender client, we are tied to the Golang version of the oldest Yocto release we support. Currently, this is "warrior", which is using an even older Golang release: 1.12. Looping in @lluiscampos, in case he has an opinion on this.

We are tied to golang release **1.14**. We did port this version of go to `zeus` and `warrior` some time ago to support Ed25519
public keys . See https://github.com/mendersoftware/meta-mender/commit/6745fd6fece4a7340bd457d495bdc2f1d3f9452e

Michaelhobo:
@kacf I expect the Go linker to strip away unused functions, so this code being in the artifacts package shouldn't change the size dramatically. I'll check to see what it looks like.

While I do that and fix up these things, can I get input on how this should get integrated into mender-artifacts? I proposed 2 ways to do CLI flags, input would be great. Also, I was thinking of modifying `getKey` to return some struct that contains `Signer`/`Verifier` instead of `[]byte`: https://github.com/mendersoftware/mender-artifact/blob/6faf8d65347bfdbec53422dd4452821c48b1ca58/cli/artifacts.go#L114

Maybe something that looks like this:

```go
type Key interface {
  Signer() artifacts.Signer
  Verifier() artifacts.Verifier
}

// OR

type Key interface {
  artifacts.Signer
  artifacts.Verifier
}
```

I haven't dug too deep yet, I'm still new to the `mender` code.

Michaelhobo:
Nope, it is definitely a lot bigger:

```bash
# New
gitpod /workspace/mender $ ls -lh mender
-rwxr-xr-x 1 gitpod gitpod 18M Nov 12 18:01 mender

# Old
gitpod /workspace/mender $ ls -lh mender
-rwxr-xr-x 1 gitpod gitpod 9.8M Nov 12 18:02 mender
```

I'll try splitting it out.

Also, it turns out go1.15+'s `ecdsa.SignASN1` just called `ecdsa.PrivateKey.Sign`, with a few default arguments. That was an easy fix!

Michaelhobo:
Ok, I think I've fixed everything now. Unfortunately, I don't know what's up with the signoff check - perhaps because I'm trying to sign off with "Michael Ho <callmemikeh@gmail.com>" instead of "Michaelhobo <callmemikeh@gmail.com>"? Mender's contributing guide says to not use aliases.

I don't know what to do about the license check. It seems like Mender has a list of license files somewhere, and the new dependencies might not be there?

kacf:
> Also, I was thinking of modifying `getKey` to return some struct that contains `Signer`/`Verifier` instead of `[]byte`:

I like that idea. Of the two you proposed, I think the bottom one is the best.

> I proposed 2 ways to do CLI flags, input would be great.
> ...
> * Add a --gcp-kms-key=name/of/my/kms/key, can migrate --key to --pem-key
> * Add a prefix to --key, like --key=gcp_kms:name/of/my/kms/key

I think I prefer option 1.

There has also been some talks lately about trying to split up various functionality into more sub commands. For example, this could be `mender-artifact gcp-sign --key=mykey` instead of `mender-artifact sign --gcp-kms-key=mykey`. Personally, I think this is overkill for this particular option, but I'd like the opinions of @lluiscampos and @oleorhagen.

> Nope, it is definitely a lot bigger:

That's been our experience in the past as well. What the compiler should leave out, and what it actually leaves out, are quite different things.

> Ok, I think I've fixed everything now. Unfortunately, I don't know what's up with the signoff check - perhaps because I'm trying to sign off with "Michael Ho [callmemikeh@gmail.com](mailto:callmemikeh@gmail.com)" instead of "Michaelhobo [callmemikeh@gmail.com](mailto:callmemikeh@gmail.com)"? Mender's contributing guide says to not use aliases.

Right. It's taken from the Linux kernel contribution guide, and is actually not completely true for Mender. What's important is that the two are matching, and that they match the author string which is tied to your Github account (since Github verifies your identity). So just change the signoff to match the author. `git commit -s` should do this automatically btw.

> I don't know what to do about the license check. It seems like Mender has a list of license files somewhere, and the new dependencies might not be there?

Yes, they are [here](https://github.com/mendersoftware/mender-artifact/blob/master/LIC_FILES_CHKSUM.sha256).

I haven't dug into the code yet, I'll do that soon.

Michaelhobo:
> > Also, I was thinking of modifying `getKey` to return some struct that contains `Signer`/`Verifier` instead of `[]byte`:
> 
> I like that idea. Of the two you proposed, I think the bottom one is the best.
> 
> > I proposed 2 ways to do CLI flags, input would be great.
> > ...
> > 
> > * Add a --gcp-kms-key=name/of/my/kms/key, can migrate --key to --pem-key
> > * Add a prefix to --key, like --key=gcp_kms:name/of/my/kms/key
> 
> I think I prefer option 1.

Thanks for the input, I'm working on the follow-up PR which should include all of that. I like the `--gcp-kms-key` flag too, because it's explicit. The only reason I bring up overloading `-key` is when thinking ahead to all the other remote key signing services. The number of key flags may continue growing, leading to flag clutter. This is easy enough to deprecate and change as the need arises.

> 
> There has also been some talks lately about trying to split up various functionality into more sub commands. For example, this could be `mender-artifact gcp-sign --key=mykey` instead of `mender-artifact sign --gcp-kms-key=mykey`. Personally, I think this is overkill for this particular option, but I'd like the opinions of @lluiscampos and @oleorhagen.

I agree, especially since `-key` is used in so many different contexts. I wouldn't want to create different subcommands for `gcp-sign`, `gcp-write-rootfs`, etc. And then again for aws, azure, and whatever others decide to add later.

It's a real shame that there isn't a standardized format for remote signing.

kacf:
> Thanks for the input, I'm working on the follow-up PR which should include all of that. I like the `--gcp-kms-key` flag too, because it's explicit. The only reason I bring up overloading `-key` is when thinking ahead to all the other remote key signing services. The number of key flags may continue growing, leading to flag clutter. This is easy enough to deprecate and change as the need arises.

Indeed, there are tradeoffs to both approaches. Another argument I found against option 2, is that different services may have different numbers of arguments for their service. A good example is the various [backend service flags for the rclone tool](https://www.mankier.com/1/rclone#Global_Flags-Backend_Flags). Yes, this results in a ton of different flags, but the key takeaway is that the flags are indeed *different*. These small differences would be difficult to express in a single string; it would not work with just a prefix.

I think this is further support for using `--gcp-kms-key`.

> I agree, especially since `-key` is used in so many different contexts. I wouldn't want to create different subcommands for `gcp-sign`, `gcp-write-rootfs`, etc. And then again for aws, azure, and whatever others decide to add later.

Agreed!

lluiscampos:
> > I proposed 2 ways to do CLI flags, input would be great.
> > ...
> > 
> > * Add a --gcp-kms-key=name/of/my/kms/key, can migrate --key to --pem-key
> > * Add a prefix to --key, like --key=gcp_kms:name/of/my/kms/key
> 
> I think I prefer option 1.
> 
> There has also been some talks lately about trying to split up various functionality into more sub commands. For example, this could be `mender-artifact gcp-sign --key=mykey` instead of `mender-artifact sign --gcp-kms-key=mykey`. Personally, I think this is overkill for this particular option, but I'd like the opinions of @lluiscampos and @oleorhagen.
> 

I support that, let's do option 1 :+1: 

We are having these "talks" but so far are just ideas in the air. Once we shape a bit of an epic and have a clear direction we could revisit all these decisions, even if it means bumping the major version of our tools.

Michaelhobo:
Thanks for your thoughts, I'll incorporate that into my next PR that makes the CLI flag changes.

Michaelhobo:
PR with the CLI flag changes and the getKey change is at https://github.com/Michaelhobo/mender-artifact-1/pull/1. Unfortunately, I don't know how to create a stacked PR from a forked repository, so the PR is showing up on my fork's PR list. Let me know if you prefer I merge that into this branch as a separate commit, or if you want to just look at the PR in my repo.

kacf:
The best would be if you could squash the changes into the existing commit and force-push this branch. There is no need to keep the history of the original code, since it was never merged. I think this fits best in one commit.

I will take a look at the details once you have done that.

Michaelhobo:
> The best would be if you could squash the changes into the existing commit and force-push this branch. There is no need to keep the history of the original code, since it was never merged. I think this fits best in one commit.
> 
> I will take a look at the details once you have done that.

It should be ready for review now. I haven't tested this end-to-end yet, because generating keys is pretty locked down, hopefully I'll get my KMS keys set up by tomorrow.

Michaelhobo:
I tested and got it working with a test key in KMS.

Michaelhobo:
@kacf when do you think you'll get around to reviewing this?

cli/artifacts_test.go
```diff
@@ -279,7 +283,7 @@ func TestArtifactsSigned(t *testing.T) {
 		"-k", filepath.Join(updateTestDir, "private.key"),
 		filepath.Join(updateTestDir, "artifact.mender")}
 	err = Run(args)
-	assert.NoError(t, err)
+	assert.Error(t, err)
```
kacf, line 286: Nice catch, I guess this should always have failed!

cli/validate_test.go
```diff
@@ -60,35 +60,62 @@ r3rtT0ysHWd7l+Kx/SUCQGlitd5RDfdHl+gKrCwhNnRG7FzRLv5YOQV81+kh7SkU
...
-	version       int
-	writeKey      []byte
-	validateKey   []byte
-	expectedError string
+var validateTests = map[string]struct {
+	version               int
+	writeKey              []byte
+	validateKey           []byte
+	expectedCreateError   string
+	expectedValidateError string
```
kacf, line 68: None of the test cases use both error strings. What is the advantage of splitting it in two?

kacf (commented):
Hey @Michaelhobo, I hadn't forgotten, but my review queue length varies every week! 🙂

This is solid work, I don't really have any other remarks than a very minor question, see below.

Michaelhobo:
> Hey @Michaelhobo, I hadn't forgotten, but my review queue length varies every week! 🙂

I didn't mean to be pushy, I'm just trying to figure out how reviews work here. At my current company, if I don't ping every day, reviews can drag on for months.

cli/validate_test.go
kacf: Ah, of course, they are after different calls. I should have caught that. Thanks for the correction.

kacf (approved):

kacf:
I'll run the full test suite now.

Michaelhobo:
I fixed the merge conflict. There also seems to be a drop in code coverage, maybe because of cli/validate.go. I added a test case, hopefully that fixes it.

kacf:
Thanks! This seems to have triggered a new unit test failure though.

Could you also rebase the changes on latest master? While this has been in progress, there has been a linting check added to the repo, and I don't think it will pass without the latest changes.

Michaelhobo:
Woops, I didn't realize `go test` wouldn't run all the tests. Fixed and it's been rebased on latest already.

kacf:
Thanks, right now the errors are due to some internal problems in our CI, just ignore them for now. We are working on it.

kacf:
Merged! Thanks @Michaelhobo!

Michaelhobo:
Awesome, thanks for the review!

PR #366, opened 2021-11-18
QA-235: pipeline: Switch to static checks by golangci-lint

lluiscampos:
And fix the code accordingly.

### DIFF
--- a/artifact/compressor_lzma.go
+++ b/artifact/compressor_lzma.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Northern.tech AS
+// Copyright 2021 Northern.tech AS
 //
 //    Licensed under the Apache License, Version 2.0 (the "License");
 //    you may not use this file except in compliance with the License.
@@ -11,6 +11,7 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
+//go:build !nolzma && cgo
 // +build !nolzma,cgo
 
 package artifact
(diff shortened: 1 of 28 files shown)


lluiscampos:
mender-test-bot start pipeline :giraffe:

lluiscampos:
@mender-test-bot start pipeline :giraffe:

utils/attributes.go
```diff
@@ -44,7 +44,6 @@ func StringsMatchingWildcards(attributes, wildcards []string) ([]string, error)
 						wildcard)
 				}
 				b.WriteString(wildcard[i : i+2])
-				i++
```
kacf, line 47: I don't think this is correct, because this case consumes two characters (btw, Github's buggy tab rendering is extremely confusing here). But there may indeed be a bug here. I think the correct change is:

```diff
-		for i := range wildcard {
+		for i := 0; i < len(wildcard); i++ {
```

kacf (changes requested):
Nice work!

artifact/compressor_lzma.go
```diff
@@ -11,18 +11,18 @@
...
+//go:build !nolzma && cgo
 // +build !nolzma,cgo
 
 package artifact
 
 import (
 	"io"
 
-	"github.com/remyoudompheng/go-liblzma"
+	xz "github.com/remyoudompheng/go-liblzma"
```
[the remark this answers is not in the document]
lluiscampos: I honestly don't know neither. The actual main package is called `xz` (see [here](https://github.com/remyoudompheng/go-liblzma/blob/master/enums.go#L6)), adding it here is just making it explicit (`gofmt` did this automatically).

utils/attributes.go
lluiscampos: I think you are right! I reverted and modified the for loop instead. I'll rerun all tests just to be sure...

lluiscampos:
@mender-test-bot start pipeline :badger:

kacf (approved):

PR #372, opened 2021-12-29
LZMA parallel compression

alpm-mobica:
Based on pxz approach - input is spited an compressed in parallel.
It causes small overhead for additional headers in the body of archive, but still it is almost 2 times smaller than gzip artifact.
Mender-artifact uses io.Copy with 32Kb buffer for data transactions for compression and totally eliminates the effect of multi-threading. It was increased to 32Mb. 
The parent repository PR is not accepted so far, so currently fork of go-lzma used.


estenberg:
Re https://tracker.mender.io/browse/MEN-2497

@kacf maybe you'd be the best to comment on this one, in particular how to handle use of forked library?

estenberg:
This (paralell lzma) feels like the right direction to go in general, and we should make lzma default if / once it is as efficient to compress as gzip. This was the sticking point for staying with gzip as the default.

alpm-mobica:
Small update, switched to lzma C lib for multi-threading, 32Mb buffer required no more.
The main issue - it takes all the cores bu default (there is custom writer with thread num setup) and could be uncomfortable for laptop users.

kacf:
Looks promising to me! I see a lot of branch updates have been happening lately, so I assume you're still polishing this?

alpm-mobica:
The main idea of latest changes was to move from hand-made threading with goroutines to stable and tested one, it is done now.
To have fine PR It worth to have an opinion of Go-expert if the sources are fine, cause so far there were no feedback from parent repository owner.

kacf:
Ok, makes sense. Let's give him a bit of time to reply.

PR #373, opened 2022-01-27
Add a new signer using Hashicorp Vault's Transit Engine

tobiasZF:
This signer signs and verifies data using Hashicorp Vault's Transit Engine.
The implementation is based on the GCP KMS signer.
A new vault-transit-key flag was added to specify the key name in Vault.

To send a signing request to Vault, i additionally need the mount path of the used transit engine. Therefore, i introduced an environment variable VAULT_MOUNT_PATH to specify this path without adding more flags to mender-artifact.

If you have a better idea on how to retrieve the transit engine's mount path from the user, feel free to comment.

Open Points:
- Add tests for new vault signer


oleorhagen:
Hello @tobiasZF 

I am sorry for the delay in review here, but we have simply not had the time to look at this until now.

I will do an initial review sometime today, then I will leave it to @kacf to do the final review, as he is more familiar with these integrations.

Thank you for your patience here.

oleorhagen:
Unfortunately, this will have to wait till monday...

kacf:
Looks promising to me!

I don't have a strong opinion about the passing of the mount path, I think an environment variable is fine for this purpose.

Indeed this will need some unit tests to get the code coverage percentage up. I might recommend turning `vaultClient` into an interface instead of a struct pointer, so you can replace it with a mock instance from the unit test.

And there are also some failures in the CI pipeline for you to look at (if something looks mysterious feel free to ask, but I'm letting you have the first look 😉).

tobiasZF:
@kacf thank you for your comments!

I will have a look at the test cases and CI failures, but this may take some time.

One question regarding the license issues:
The vault go package is released under MPL-2.0, which is not (yet) included in your `LIC_FILES_CHKSUM.sha256` file. Is this license compatible to your project/ license requirements?

kacf:
That's a very good question @tobiasZF, I had not looked at the licenses yet. I think MPL should be fine, there is some fine print, but in essence it has the same requirements as LGPL-2.1, which we already link to.

So just add it to the checksum file under a new header.

tobiasZF:
@kacf I have added the missing test cases and resolved all failures in the CI pipeline.
The most current version of vault/api go package (v1.3.1) is not compatible with go 1.14. That's why i needed to downgrade the vault/api package to v1.2.0.

If you upgrade the overall go version of this project in future, we can use the most current version of the vault/api client :-)

I also added support for using the key rotation feature of Hashicorp Vault.

kacf:
Thanks @tobiasZF, I'll take a look shortly!

> The most current version of vault/api go package (v1.3.1) is not compatible with go 1.14. That's why i needed to downgrade the vault/api package to v1.2.0.
> 
> If you upgrade the overall go version of this project in future, we can use the most current version of the vault/api client :-)

The reason we use it is because of the Yocto Dunfell LTS release, where the version of Go is 1.14. After April this branch will go into maintenance mode, so we may revisit the version then.

kacf (changes requested):
Nice work on the tests! I've got a couple of suspicious cast operations pointed out below, but apart from that this looks very good.

tobiasZF:
@kacf thank you for your review and comments, all valid points.
I have tried to rework the lines you mentioned, may you have a look again if this resolves your comments?

kacf (approved):
Thanks, very thorough! I'll run this through our full test suite next.

kacf:
Thank you very much @tobiasZF!

PR #413, opened 2022-08-18
feat(PKCS#11): Add PKCS#11 standard support when signing artifacts

MaciejTe:
feat(PKCS#11): Add PKCS#11 standard support when signing artifacts

Changelog: Title
Ticket: MEN-5759
Signed-off-by: Maciej Tomczuk <maciej.tomczuk@northern.tech>


oleorhagen (approved):
Nice!

artifact/signer.go
```diff
@@ -302,3 +308,67 @@ func GetKeyAndSignMethod(keyPEM []byte) (*SigningMethod, error) {
...
 	return nil, errors.Wrap(err, "signer: unsupported private key type or error occured")
 }
+
+type PKCS11Signer struct {
+	Key openssl.PrivateKey
+}
+
+func NewPKCS11Signer(pkcsKey string) (*PKCS11Signer, error) {
+	if len(pkcsKey) == 0 {
+		return nil, errors.New("PKCS signer: missing key")
```
tranchitella, line 318:
```suggestion
		return nil, errors.New("PKCS#11 signer: missing key")
```

MaciejTe: done!

artifact/signer.go
```diff
@@ -302,3 +308,67 @@ func GetKeyAndSignMethod(keyPEM []byte) (*SigningMethod, error) {
...
+	}
+	const engineId = "pkcs11"
+
+	key, err := loadPrivateKey(string(pkcsKey), engineId)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
MaciejTe: done!

artifact/signer.go
```diff
@@ -302,3 +308,67 @@ func GetKeyAndSignMethod(keyPEM []byte) (*SigningMethod, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		key, err = engineLoadPrivateKeyFunc(engine, keyFile)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		return nil, errors.New("PKCS#11 URL prefix not found")
```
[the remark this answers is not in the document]
MaciejTe: done!

.gitlab-ci.yml
```diff
@@ -5,16 +5,16 @@ stages:
...
+    file: ".gitlab-ci-check-golang-lint.yml"
+  - project: "Northern.tech/Mender/mendertesting"
+    file: ".gitlab-ci-github-status-updates.yml"
+  - project: "Northern.tech/Mender/mendertesting"
+    file: ".gitlab-ci-check-commits.yml"
+  - project: "Northern.tech/Mender/mendertesting"
+    file: ".gitlab-ci-check-license.yml"
 
-image: golang:1.14-alpine3.11
+image: golang:1.18-alpine3.16
```
lluiscampos, line 17: As discussed in person, please move all non-feature related changes to separate commits. At least this one could go on its own (chore: update golang version) and the yaml formatting could also go on their own.

Makefile
```diff
@@ -43,31 +43,48 @@ build:
...
+	image_id=$$(docker build -f Dockerfile.binaries -q --build-arg PLATFORM=darwin . ) && \
+	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact* /binary" && \
+	docker image rm $$image_id
+
+build-natives-contained-windows:
+	rm -f mender-artifact-windows.exe && \
+	image_id=$$(docker build -f Dockerfile.binaries -q --build-arg PLATFORM=windows . ) && \
+	docker run --rm --entrypoint "/bin/sh" -v $(shell pwd):/binary $$image_id -c "cp /go/bin/mender-artifact* /binary" && \
+	docker image rm $$image_id
+
```
lluiscampos, line 86: Can we revert all this? I don't really understand the purpose (we already have a mechanism to cross-compile, right?). I suspect that changing this will trigger problems in mender-qa and/or meta-mender.

lluiscampos: by "all this" I mean changes in files `Makefile`, `Dockerfile.binaries` and `gitlab-ci.yml` (the parts related to the new make targets).

MaciejTe: I did not understand why all binaries were built in one stage - a rule of thumb is to split operations in CI and don't group them unnecessarily. From my perspective, it's better to see which platform builds correctly and which is not right away.

MaciejTe: As discussed, I'll create a separate PR with mentioned CI refactor.

Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM golang:1.14 as builder
-RUN apt-get update && apt-get install -qyy liblzma-dev
+FROM golang:1.17 as builder
+RUN apt-get update && apt-get install -qyy liblzma-dev libssl-dev
```
mzedel, line 2: https://github.com/mendersoftware/mender-artifact/pull/287 <- not sure how relevant that discussion is by now @kacf @lluiscampos @oleorhagen, in any case this should be aligned with the `Dockerfile.binaries`

Makefile
mzedel: > a rule of thumb is to split operations in CI and don't group them unnecessarily. From my perspective, it's better to see which platform builds correctly and which is not right away.

The build is one step in CI and should work at all times for all platforms, if any platform build fails you will still see this in the job + you would see if a platform fails in the following tests.
By building all platforms in one step you can use the cross compilation benefit that go as a language brings and reuse the build environment.

.gitlab-ci.yml
MaciejTe: done!

Dockerfile
MaciejTe: done, thanks!

lluiscampos:
@mender-test-bot start pipeline :pizza:

lluiscampos:
@mender-test-bot start pipeline --pr meta-mender/1761 :moon_cake:

lluiscampos (approved):

PR #414, opened 2022-08-19
Fix/docker version compatibility + docker image execution

mzedel:


mzedel:
non buildkit using pipeline: https://gitlab.com/Northern.tech/Mender/mender-artifact/-/pipelines/617868677
buildkit using pipeline: https://gitlab.com/Northern.tech/Mender/mender-artifact/-/pipelines/617871038

mzedel:
> @mender-test-bot, Let me know if you want to start the integration pipeline by mentioning me and the command "start pipeline".
> 
> my commands and options

🧱

.gitlab-ci.yml
```diff
@@ -13,6 +13,8 @@ include:
     file: '.gitlab-ci-check-commits.yml'
   - project: 'Northern.tech/Mender/mendertesting'
     file: '.gitlab-ci-check-license.yml'
+  - project: 'Northern.tech/Mender/mendertesting'
+    file: '.gitlab-ci-check-docker-build.yml'
```
lluiscampos, line 17: Why including the template if we are overriding `build:docker` ?

Makefile
```diff
@@ -58,13 +58,13 @@ build-natives:
 
 build-contained:
 	rm -f mender-artifact && \
-	image_id=$$(docker build -f Dockerfile . | awk '/Successfully built/{print $$NF;}') && \
+	image_id=$$(docker build -f Dockerfile -q .) && \
```
lluiscampos, line 61: As the `Makefile` can be used outside of our CI, I think we should still support building without BUILDKIT. Can we check for it somehow in this command to then conditionally use `awk`?

mzedel: => https://github.com/mendersoftware/mender-artifact/pull/414#issuecomment-1220657061 - this works for both...

.gitlab-ci.yml
mzedel: argh sorry, that's a leftover - my intent was to use the standardized approach, however the `build:docker` job seems to not be used anywhere in the pipeline, I forgot to remove the template (preferably I would also remove the `build:docker` job, but I'm not sure if you might want to keep it to just test if that `make` target still works)

Makefile
lluiscampos: Sorry :facepalm:

.gitlab-ci.yml
lluiscampos: While I see your good intentions, I think for this case is better to use the `Makefile` from the CI to guarantee we don't break it :+1:

lluiscampos (approved):

PR #415, opened 2022-08-21
3.6.x

tranchitella:


tranchitella:
Required by https://github.com/mendersoftware/integration/pull/2100

PR #422, opened 2022-09-12
Sign with HSM test.

merlin-northern:
Changelog: Title
Ticket: MEN-5761

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -96,8 +96,10 @@ test:smoketests:linux:
   image: golang:1.18-buster
   before_script:
     - apt-get update && apt-get install -q -y make liblzma-dev libssl-dev
+    - bash tests/test_sign_with_hsm/test_sign_with_hsm.sh --setup
   script:
     - touch test.txt
+    - TEST_MENDER_ARTIFACT_PATH=./mender-artifact-linux bash tests/test_sign_with_hsm/test_sign_with_hsm.sh
     - ./mender-artifact-linux
     - ./mender-artifact-linux --version
     - ./mender-artifact-linux write module-image -t test -o test.mender -T script -n test -f test.txt
--- a/tests/test_sign_with_hsm/config.sh
+++ b/tests/test_sign_with_hsm/config.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+# Copyright 2022 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+declare -A TEST_CONFIG=(
+    [sopin]="0002"
+    [pin]="0001"
+    [keylen]="2048"
+    [privatekey_path]="/tmp/private.key"
+    [artifact]="/tmp/ci-artifact.mender"
+    [mender_artifact]="${TEST_MENDER_ARTIFACT_PATH}"
+)
+
+TEST_NAME="test_sign_with_hsm"
+
+TEST_CONFIGURED=1
--- a/tests/test_sign_with_hsm/functions.sh
+++ b/tests/test_sign_with_hsm/functions.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+# Copyright 2022 Northern.tech AS
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+set -ex
+
+function log() {
+    echo "$(date) ${TEST_NAME:-"unknown"}/$$ $HOSTNAME $@"
+}
+
+function is_debian() {
+    [[ "$(lsb_release -d 2> /dev/null)" =~ "Debian" ]] || [[ "$(cat /etc/issue 2> /dev/null)" =~ "Debian" ]]
+}
+
+function debian_setup() {
+    local pin="${TEST_CONFIG[pin]}"
+    local sopin="${TEST_CONFIG[sopin]}"
+
+    log "running setup for debian"
+    apt-get update
+    apt-get install -qy softhsm2 softhsm2-common libsofthsm2 libengine-pkcs11-openssl opensc-pkcs11 opensc gnutls-bin openssl gawk
+    echo "module: /usr/lib/softhsm/libsofthsm2.so" > /usr/share/p11-kit/modules/softhsm2.module
+    mkdir -p /softhsm/tokens
+    echo "directories.tokendir = /softhsm/tokens" > /softhsm/softhsm2.conf
+    export SOFTHSM2_CONF=/softhsm/softhsm2.conf
+    softhsm2-util --init-token --free --label unittoken1 --pin "$pin" --so-pin "$sopin"
+    openssl genrsa -out "${TEST_CONFIG[privatekey_path]}" "${TEST_CONFIG[keylen]}"
+    pkcs11-tool --module /usr/lib/softhsm/libsofthsm2.so --login --pin "$pin" --write-object "${TEST_CONFIG[privatekey_path]}" --type privkey --id 0909 --label privatekey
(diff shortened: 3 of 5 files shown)


merlin-northern:
this one is a copy of https://github.com/mendersoftware/mendertesting/pull/214 to show that it runs and passes.

merlin-northern:
I am not running the full pipeline as the patch goes only in the tests and local pipeline.

kjaskiewiczz (approved):
looks good, thanks!

tests/test_sign_with_hsm/get-and-run.sh
```diff
@@ -0,0 +1,23 @@
...
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+set -ex
+
+root_directory=$(mktemp -d)
+cd "$root_directory"
+for i in config.sh functions.sh setup.sh test_sign_with_hsm.sh; do
+    curl https://raw.githubusercontent.com/mendersoftware/mender-artifact/master/tests/test_sign_with_hsm/"$i" -o "$i"
```
lluiscampos, line 21: This is very dangerous: a PR modifying the test will never be able to be tested in CI, as this line is getting latest master regardless.

(it is unused, anyway)

tests/test_sign_with_hsm/get-and-run.sh
```diff
@@ -0,0 +1,23 @@
...
+done
+bash test_sign_with_hsm.sh
```
lluiscampos, line 23: This script is a leftover from the original design in `mendertesting`. Please remove.

tests/test_sign_with_hsm/functions.sh
```diff
@@ -0,0 +1,54 @@
...
+    local sopin="${TEST_CONFIG[sopin]}"
+    local artifact="${TEST_CONFIG[artifact]}"
+
+    [[ "$TEST_CONFIGURED" == "1" ]] || {
+        log "test is not configured"
+        return 1
+    }
+    "${TEST_CONFIG[mender_artifact]}" write module-image -T local-type -n ci-tests-artifact-1 -t ci-type-1 -o "$artifact"
+    echo -ne "[openssl_init]\nengines=engine_section\n\n[engine_section]\npkcs11 = pkcs11_section\n\n[pkcs11_section]\nengine_id = pkcs11\nMODULE_PATH = /usr/lib/softhsm/libsofthsm2.so\ninit = 0\n" >> /etc/ssl/openssl.cnf
+    p11tool --login --provider=/usr/lib/softhsm/libsofthsm2.so --set-pin="$pin" --list-all-privkeys | awk -v pin="$pin" -v menderartifact="${TEST_CONFIG[mender_artifact]}" -v artifact="$artifact" '/URL/{ rc=system(menderartifact" sign --key-pkcs11 \""$NF";pin-value="pin"\" "artifact); exit(rc); }' && log PASSED
```
lluiscampos, line 53: Could you please have this as a multi-line script? It gets very difficult to read.

lluiscampos, line 53: Should we also verify `mender-artifact read`?

.gitlab-ci.yml
```diff
@@ -98,6 +98,7 @@ test:smoketests:linux:
     - apt-get update && apt-get install -q -y make liblzma-dev libssl-dev
   script:
```
lluiscampos, line 99: The dependencies should be installed in `before_script`.

Rationale: if I am running this test locally, I most likely want to install the dependencies once (iow, call `debian_setup`) but run the test many times (iow `test_main`).

tests/test_sign_with_hsm/test_sign_with_hsm.sh
```diff
@@ -0,0 +1,27 @@
...
+[[ -f ${TEST_ROOT}/functions.sh ]] && . ${TEST_ROOT}/functions.sh
+[[ -f ${TEST_ROOT}/config.sh ]] && . ${TEST_ROOT}/config.sh
+
+is_debian || {
+    log "SKIPPING under non Linux/Debian."
+    exit 0
+}
+
+debian_setup
+test_main
```
lluiscampos, line 27: In my opinion, the whole design looks over-engineered. I would be happy with everything in one single `.sh` file (or maybe setup in one file, the rest in another file). Can you elaborate the benefits of having config, functions, setup, and then this test?

tests/test_sign_with_hsm/config.sh
```diff
@@ -0,0 +1,27 @@
...
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+declare -A TEST_CONFIG=(
+    [sopin]="0002"
+    [pin]="0001"
+    [keylen]="2048"
+    [privatekey_path]="/tmp/private.key"
+    [artifact]="/tmp/ci-artifact.mender"
+    [mender_artifact]="${TEST_MENDER_ARTIFACT_PATH}"
```
lluiscampos, line 22: This I like very much :rocket: as it will allow us easily to replace it with an instrumented binary to collect code coverage.

tests/test_sign_with_hsm/setup.sh
```diff
@@ -0,0 +1,20 @@
...
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+set -ex
+
+export TEST_ROOT=$(dirname "$0")
+[[ -f ${TEST_ROOT}/functions.sh ]] && . ${TEST_ROOT}/functions.sh
+
+is_debian && debian_setup
```
lluiscampos, line 20: I suggest we completely remove the `is_debian` test. I don't want to take any risk in the test not being run and going unnoticed. It should only be called by CI on a debian based image... so as far as the test is concern it should just run and fail if something doesn't work.

tests/test_sign_with_hsm/functions.sh
lluiscampos, line 53: Here there is a mix of global variables (i.e. `${TEST_CONFIG[mender_artifact]}`) and local ones (i.e. `$artifact`, `$pin`). Please make it consistent.

lluiscampos (changes requested):
Looking good! I think it can get a bit more polishing :soap:

tests/test_sign_with_hsm/get-and-run.sh
merlin-northern: it was a helper, to run locally. removed.

tests/test_sign_with_hsm/get-and-run.sh
merlin-northern: removed -- I kept it just in case someone would need to get&run.

tests/test_sign_with_hsm/functions.sh
merlin-northern: sure. done.

merlin-northern: there apparently is no support for `--key-pkcs11` in read.

.gitlab-ci.yml
merlin-northern: sure. modified.

tests/test_sign_with_hsm/test_sign_with_hsm.sh
merlin-northern: of course: this is only a beginning: what I have in mind is a framework for those kind of tests. Could we please keep it? It does not bother to have a clean separation between the stages, right?

tests/test_sign_with_hsm/config.sh
merlin-northern: great!

tests/test_sign_with_hsm/setup.sh
merlin-northern: I added an explicit `exit 1` in case we are running on non Debian. I want to keep `is_debian` logic, if that does not bother you. the test will now fail as you suggest.

tests/test_sign_with_hsm/functions.sh
merlin-northern: I moved `${TEST_CONFIG[mender_artifact]}` to a local too.

tests/test_sign_with_hsm/get-and-run.sh
lluiscampos: I see. As a general idea having a general wrapper I am not against. This is for example what we ended up doing in mender-docs to have an easy way to run the tests locally and on CI. But it should 1) be used in CI and 2) not curl anything from upstream but rather use the local files in the repo.

Anyway, if you removed it there is no much more to discuss :sweat_smile:

tests/test_sign_with_hsm/functions.sh
lluiscampos: Okey (that was unexpected, though). Can we at least `read` it with no key and get the info? It will then say that the signature cannot be validated, but that is good enough I think.

lluiscampos (approved):

tests/test_sign_with_hsm/functions.sh
merlin-northern: [sure thing](https://github.com/mendersoftware/mender-artifact/pull/422/files#diff-510ab35d2b35be306594c085dc4d5ff34bbc6738893bde54d59657fe4d0b5e50R81) I added read with a check for `Signature: signed`

lluiscampos (approved):

tests/test_sign_with_hsm/functions.sh
merlin-northern: And thanks a lot for that comment. it completes the test. thanks.

PR #425, opened 2022-09-15
chore: Cancel changelog 19e706e

lluiscampos:
Tests do not honor a changelog entry.

Cancel-changelog: 19e706ea441e0b24d50fdad95cb4ccc5851eabe0


merlin-northern (approved):
sorry for that. there must be a mistake in my script, I expilictly test for test: in the message.

PR #431, opened 2022-09-26
test: verify artifacts signed with PKCS#11 using file-based public key

tranchitella:
Ticket: MEN-5933
Changelog: None

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


lluiscampos (approved):
Splendid.

PR #434, opened 2022-10-10
MEN-5833: Collect code coverage from PKCS#11 test

0lmi:
Changelog: None
Ticket: MEN-5833
Signed-off-by: Alex Miliukov <oleksandr.miliukov@northern.tech>


0lmi:

main_test.go
```diff
@@ -12,4 +12,78 @@
...
+	"os"
+	"os/signal"
+	"testing"
+
+	"github.com/mendersoftware/go-lib-micro/log"
+)
+
+func init() {
+	// Make sure main does not exit before we have gathered coverage.
+	log.Log.ExitFunc = func(int) {}
```
oleorhagen, line 36: Question: Why do we need `go-lib-micro` here?

oleorhagen (commented):
It looks to me like we can solve this using our already existing binary instrumentation for collecting coverage.

If you want me to give you a quick run-through of this, just ping me :)

oleorhagen (changes requested):
After a quick f2f we decided the already existing binary-instrumentation for test-coverage is plenty here

0lmi:
Alternative implementation using `gobinarycoverage` https://github.com/mendersoftware/mender-artifact/pull/435

0lmi:
@oleorhagen if to be honest I think this approach is more unified than using `gobinarycoverage` due to it's used everywhere else (mender-cli, all backend services...). Obviously the result is the same, but unification is valuable for the team.

oleorhagen:
> @oleorhagen if to be honest I think this approach is more unified than using `gobinarycoverage` due to it's used everywhere else (mender-cli, all backend services...). Obviously the result is the same, but unification is valuable for the team.

You can make the decision here yourself. 

I would favour the gobinarycoverage approach, because it's already in-place, and we would not duplicate functionality.

lluiscampos:
Close in favor of #435?

PR #442, opened 2022-10-26
chore: apk requirements to run mender-artifact on alpine

0lmi:
Changelog: None
Ticket: MEN-5906
Signed-off-by: Alex Miliukov <oleksandr.miliukov@northern.tech>

### DIFF
--- a/apk-run-requirements.txt
+++ b/apk-run-requirements.txt
@@ -1,0 +1,3 @@
+libc6-compat
+xz
+libssl1.1


apk-requirements.txt
```diff
@@ -0,0 +1,2 @@
+libc6-compat
+xz-dev
```
lluiscampos, line 2: We need `xz` instead of `xz-dev`, right? These are supposed to be runtime requirements?

lluiscampos, line 2: We are also missing libssl. It is installed by default in Alpine Linux, but I think is good to be explicit (specially if they change to libssl3 any time soon).
```suggestion
xz-dev
libssl1.1
```

lluiscampos (changes requested):

apk-requirements.txt
0lmi: I don't really know those details. Let me check...

0lmi: Yep, you are totally right 👍

0lmi: 👍

lluiscampos (approved):

0lmi:

PR #503, opened 2023-06-29
Add automation to combine dependabot PRs

lluiscampos:


.github/stale.yml
```diff
@@ -2,10 +2,6 @@
 daysUntilStale: 60
 # Number of days of inactivity before a stale issue is closed
 daysUntilClose: 7
-# Issues with these labels will never be considered stale
-# exemptLabels:
- # - pinned
```
mzedel, line 7: Would it make sense to apply the same `  - dependencies` exemption here as in `mender-cli`?

mzedel (approved):
Thanks for this! I wasn't sure about the non-backend repos considering the likely conflicts in the vendored dependencies - but we'll see soon how it works! 🤞

.github/stale.yml
lluiscampos: It is already there, line 15 :+1: I am just removing the "duplicate".

mzedel: argh sorry for the noise, I just looked at the github diff 🙈

lluiscampos:
> Thanks for this! I wasn't sure about the non-backend repos considering the likely conflicts in the vendored dependencies - but we'll see soon how it works! crossed_fingers

Yes. I am not 100% sure neither. This repository is the best candidate to try it out IMO :smiley:

PR #513, opened 2023-07-11
fix: add compilation instructions to README

TheYoctoJester:
Changelog: Title
Ticket: None

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,46 @@ For sample usage, please see the [Mender client source code](https://github.com/
 You can find the latest `mender-artifact` binaries in the [Downloads page on
 Mender Docs](https://docs.mender.io/downloads).
 
+## Building from source
+
+**Note: The build process is only supported on Linux-based environments.**
+
+### Containerized
+
+The preferred way to build `mender-artifact` from source is containerized in `docker`.
+This process can be used to generate binaries suitable for running on Linux, Windows and MacOS
+hosts. 
+
+You need the following prerequisites:
+- docker [Installation instructions](https://docs.docker.com/get-docker/)
+- the `make` tool. Most distributions offer a `make` package. Install it using your native
+  package management tool, such as `apt`, `yum`, `pacman`, ...
+
+In the source directory, build `mender-artifact` by issueing
+```
+make build-natives-contained
+```
+
+This results in the self-contained binary `mender-artifact`. You can leave it in place,
+or move it to a location on your systems `PATH`.
+
+**Note: containerized building on non-x86 platform is not supported yet**
+
+### Native
+
+To build `mender-artifact` from source you need the following prerequisites:
+(packages given for Debian-based environments)
+- the Go programming language. [Installation instructions](https://go.dev/doc/install)
+- the packages `build-essential liblzma-dev libssl-dev pkg-config`
+
+In the source directory, `mender-artifact` is built by simply issueing
+```
+make
+```
+
+This results in the self-contained binary `mender-artifact`. You can leave it in place,
+or move it to a location on your systems `PATH`.
+
 ## Enabling auto-completion in Bash & ZSH
 
 ### Automatic installation through the Makefile


README.md
```diff
@@ -55,6 +55,20 @@ For sample usage, please see the [Mender client source code](https://github.com/
 You can find the latest `mender-artifact` binaries in the [Downloads page on
 Mender Docs](https://docs.mender.io/downloads).
 
+## Building from source
+
+To build `mender-artifact` from source you need the following prerequisites:
+- the Go programming language. (Installation instructions](https://go.dev/doc/install)
+- on Ubuntu, the packages `build-essential liblzma-dev libssl-dev pkg-config`
```
mzedel, line 62:
```suggestion
To build `mender-artifact` from source you need the following prerequisites:
- docker
```
the above instructions may not work on non-linux devices, whereas the docker build does

TheYoctoJester: @mzedel better that?

mzedel: In CI we do [this](https://github.com/mendersoftware/mender-artifact/blob/master/.gitlab-ci.yml#L59C35-L59C35)
 - so `make build-natives-contained` and you get native binaries for linux, mac & windows (although support for mac & windows is "extra light") right away, same as the binaries we publish & test - OTOH we don't test the result of the bare `make` at all...
+ if we ever need other dependencies, we won't have to adjust this part of the readme (which is IMO a change that tends to be overlooked)...

TheYoctoJester: @mzedel just tried to test and verify it. Sadly:
```
make build-natives-contained
rm -f mender-artifact-darwin mender-artifact-linux mender-artifact-windows.exe && \
	image_id=$(docker build -f Dockerfile.binaries -q .) && \
	docker run --rm --entrypoint "/bin/sh" -v /Users/josef/Projects/mender/mender-artifact:/binary $image_id -c "cp /go/bin/mender-artifact* /binary" && \
	docker image rm $image_id
[+] Building 0.0s (9/14)
 => [internal] load build definition from Dockerfile.binaries                                                    0.0s
 => => transferring dockerfile: 83B                                                                              0.0s
 => [internal] load .dockerignore                                                                                0.0s
 => => transferring context: 2B                                                                                  0.0s
 => [internal] load metadata for docker.io/library/alpine:3.17.2                                                 0.5s
 => [internal] load metadata for docker.io/library/golang:1.19.5-bullseye                                        0.5s
 => [internal] load build context                                                                                0.1s
 => => transferring context: 456.06kB                                                                            0.1s
 => CACHED [builder 1/6] FROM docker.io/library/golang:1.19.5-bullseye@sha256:78b1726316ebfdec022fe7628f07547ca  0.0s
 => CACHED [stage-1 1/3] FROM docker.io/library/alpine:3.17.2@sha256:ff6bdca1701f3a8a67e328815ff2346b0e4067d32e  0.0s
 => ERROR [builder 2/6] RUN apt-get update &&     apt-get install -y     gcc gcc-mingw-w64 gcc-multilib     git  3.7s
 => [stage-1 2/3] RUN apk add --no-cache libc6-compat xz-dev libressl-dev                                        3.6s
------
 > [builder 2/6] RUN apt-get update &&     apt-get install -y     gcc gcc-mingw-w64 gcc-multilib     git make     musl-dev liblzma-dev libssl-dev:
#8 0.344 Get:1 http://deb.debian.org/debian bullseye InRelease [116 kB]
#8 0.477 Get:2 http://deb.debian.org/debian-security bullseye-security InRelease [48.4 kB]
#8 0.536 Get:3 http://deb.debian.org/debian bullseye-updates InRelease [44.1 kB]
#8 0.587 Get:4 http://deb.debian.org/debian bullseye/main arm64 Packages [8071 kB]
#8 2.265 Get:5 http://deb.debian.org/debian-security bullseye-security/main arm64 Packages [248 kB]
#8 2.324 Get:6 http://deb.debian.org/debian bullseye-updates/main arm64 Packages [12.1 kB]
#8 2.935 Fetched 8539 kB in 3s (3160 kB/s)
#8 2.935 Reading package lists...
#8 3.234 Reading package lists...
#8 3.537 Building dependency tree...
#8 3.636 Reading state information...
#8 3.648 Package gcc-multilib is not available, but is referred to by another package.
#8 3.648 This may mean that the package is missing, has been obsoleted, or
#8 3.648 is only available from another source
#8 3.648
#8 3.690 E: Package 'gcc-multilib' has no installation candidate
------
executor failed running [/bin/sh -c apt-get update &&     apt-get install -y     gcc gcc-mingw-w64 gcc-multilib     git make     musl-dev liblzma-dev libssl-dev]: exit code: 100
make: *** [build-natives-contained] Error 1
```

mzedel: that seems to be the unreliable package situation on debian (package doesn't exist for the arm platform)... adding a `--platform=linux/amd64` to the docker build fixes this - otherwise a [CI build](https://gitlab.com/Northern.tech/Mender/mender-artifact/-/jobs/4637681841) (where we don't build for arm yet) ran without problems...

mzedel: to me the focus on Ubuntu is ignoring what our user base looks like - the majority of our users is not working on linux (but working with), so IMO we should make sure that the instructions we provide reflect that (especially for tools that are likely to be run on a dev machine instead of on a target device).

As to the change here: `docker.io` [is actively discouraged](https://docs.docker.com/engine/install/ubuntu/#uninstall-old-versions) - IMO linking to https://docs.docker.com/get-docker/ should be enough and OS independent.

TheYoctoJester: @mzedel once we are talking about building root filesystems, people almost certainly are working on Linux. At least from what I can tell, the common concensus there is "provide instructions for debianoids, let others figure out the rest themselves".
I'm okay with referring to the official Docker docs, and wording it less Ubuntu-centric.

mzedel: on this I'm limited to what I see as questions on Mender Hub, PRs here and in the HM analytics - they point to a good chunk (in case of the analytics a majority) of non-linux users - if you have data that says otherwise, feel free to disregard this entirely, but my experience says we lack support for non-linux platform in the tools we provide (and I think we should be mindful not to further this path).

TheMeaningfulEngineer (changes requested):
I appreciate the effort of trying to expand the readme, but I don't agree with setting expectations which we likely won't be fulfilling.

I would limit this to only to the steps which we build in the testing pipeline.
So probably only Linux down to the distro on which we actually build.

Everything else has a chance to become outdated.

My reasoning is it's better to have less things in the docs that work than cover edge cases that won't be maintained.
The edge cases can be a hub post.

mzedel:
> I would limit this to only to the steps which we build in the testing pipeline. So probably only Linux down to the distro on which we actually build.
@TheMeaningfulEngineer => https://gitlab.com/Northern.tech/Mender/mender-artifact/-/jobs/4646331569/artifacts/browse

TheYoctoJester:
Additionally, just tried on Windows
```
sepp@office-windows MINGW64 ~/Documents/Projects/m/mender-artifact (master)
$ make build-natives-contained
rm -f mender-artifact-darwin mender-artifact-linux mender-artifact-windows.exe && \
image_id=$(docker build -f Dockerfile.binaries -q .) && \
docker run --rm --entrypoint "/bin/sh" -v /c/Users/sepp/Documents/Projects/m/mender-artifact:/binary $image_id -c "cp /go/bin/mender-artifact* /binary" && \
docker image rm $image_id
Unable to find image 'failed:latest' locally
docker: Error response from daemon: pull access denied for failed, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
See 'docker run --help'.
make: *** [Makefile:74: build-natives-contained] Error 125
```
So even after jumping through the hoops of getting `make`, `git`  and `docker` set up on Windows X86, it falls over.

The desire for going beyond Linux is understandable, but at least in the current state of things not something we should encourage.

TheYoctoJester:
@mender-test-bot start pipeline please

merlin-northern:
I merged as Josef asked.

PR #514, opened 2023-07-13
fix: Reader cannot parse files with zst compression

alfrunes:


kacf (approved):
Thanks, change looks good to me!

Please don't use `--fast` option on client pipelines, though. It omits nearly all client tests.

kjaskiewiczz:
@alfrunes looks good; look at the comment [here](https://northerntech.atlassian.net/browse/MEN-6617?focusedCommentId=98133) ;)

kjaskiewiczz:
https://github.com/mendersoftware/mender-artifact/pull/515 @alfrunes same as yours (almost), but with green basic pipeline and running full pipeline (without --fast option)

alfrunes:
> Please don't use `--fast` option on client pipelines, though. It omits nearly all client tests.

I used the fast option because this code path clearly is not tested.

kacf:
> I used the fast option because this code path clearly is not tested.

Ok, fair point. 🙂

PR #517, opened 2023-07-19
Merge master into 3.10.x

tranchitella:


tranchitella:
Merging without running the full pipeline, this release branch is not in use (yet) and we'll run full integration tests as part of the release.

kacf:
This unfortunately needs to reverted, because of https://northerntech.atlassian.net/browse/MEN-6449. We cannot include the dependabot updates in the 3.10.1 patch release.

PR #522, opened 2023-08-01
fix(crypto): Support signing with keys in the PKCS8 format

oleorhagen:
Ticket: None
Changelog: Support extracting RSA keys for signing artifacts from the PKCSv8 format.


oleorhagen:
Nvm, I should pull more often

PR #523, opened 2023-08-03
chore: Remove action to combine PRs

lluiscampos:
It turned out to be of little help in this repository

This reverts commits:
* 44c1183acd739f17fd1d1d2199ad71ce7498010c
* d0268bbe0dcf16f418335dae867ee36da1678bb1


mzedel (approved):
IMO the reason why this is not really helpful is rather the vendor folder, as most of the backend repos suffer from similar combinations - but that's unlikely to change here in the near future I think...

PR #542, opened 2023-09-16
Add arm64 docker build

oldgiova:
Ticket: QA-613

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,7 +26,6 @@ cache:
     - /go/src/gopkg.in
 
 variables:
-  DOCKER_HOST: tcp://docker:2375/
   DOCKER_REPOSITORY: mendersoftware/mender-artifact
   S3_BUCKET_NAME: "mender"
   S3_BUCKET_PATH: "mender-artifact"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -71,6 +70,34 @@ build:make:
     DOCKER_TLS_VERIFY: "1"
     DOCKER_TLS_CERTDIR: "/certs"
 
+build:docker-multiplatform:
```
lluiscampos, line 73: I am conflicted about this, can't we [cross-compile](https://github.com/mendersoftware/mender-artifact/blob/5f07d465fd20b9b843c84966d4f277d4c484c6c7/Makefile#L65) as we do for the other platforms? What is the benefit of native build in a container?

oldgiova: Here my need is just to produce a `mender-artifact-linux-arm64` artifact to be consumed [here](https://github.com/mendersoftware/generate-delta-worker/pull/44/files#diff-037ea159eb0a7cb0ac23b851e66bee30fb838ee8d0d99fa331a1ba65283d37f7R81).

This is the quickest way to achieve the goal without refactoring the whole process. If we want to do the best things - full cross-compile, also for the other platforms - then I need a lot of time more.

May I suggest a follow-up maybe?

lluiscampos: But that is my point: we already have the Makefile that cross-compiles for windows and MacOS. The task would be to add a linux arm64 target and add it to `build-natives`. Let me give it a quick try and see it it would work for you.

PR coming up.

lluiscampos: @oldgiova I think this one should achieve the same thing. Will this binary work for you?
* https://github.com/mendersoftware/mender-artifact/pull/544

oldgiova: The key difference is the current pipeline is working cross-OS not cross-platform; when it comes to a amd/arm cross compile, then we need `docker buildx` either with cross-compilation mode (difficult to achieve, faster builds) or emulation mode (easy to achieve, slower builds).
Some reference: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

I think just adding the arm64 reference won't work without changing also the build method, unless we go for emulation mode; but emulation mode is way slower, and in this PR I just added it for the arm64 build only, to not slow down too much the build pipeline.

lluiscampos: You are totally right. Thank you.

lluiscampos (approved):

lluiscampos:

lluiscampos:
* 3.10.x
* 3.9.x

PR #546, opened 2023-09-22
3.9.x multiplatform build

oldgiova:
Ticket: QA-613

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,7 +24,6 @@ cache:
     - /go/src/gopkg.in
 
 variables:
-  DOCKER_HOST: tcp://docker:2375/
   DOCKER_REPOSITORY: mendersoftware/mender-artifact
   S3_BUCKET_NAME: "mender"
   S3_BUCKET_PATH: "mender-artifact"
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oldgiova `mender-artifact` 3.9.x is not supported anymore. It is 3.10 and 3.8. For 3.8 I am not sure about the value of porting this change, as it will be EOL by December of this year (and very unlikely that we do any release of the tool before that)

oldgiova:
> @oldgiova `mender-artifact` 3.9.x is not supported anymore. It is 3.10 and 3.8. For 3.8 I am not sure about the value of porting this change, as it will be EOL by December of this year (and very unlikely that we do any release of the tool before that)

That's unfortunate... My intention was to push the 3.9.0 build to add arm support to the current [generate-delta-worker pipeline](https://github.com/mendersoftware/generate-delta-worker/blob/master/.gitlab-ci.yml#L3) which uses the `3.9.0`.
So now I guess we have to move the generate-delta-worker to `3.10.1`?

lluiscampos:
> > @oldgiova `mender-artifact` 3.9.x is not supported anymore. It is 3.10 and 3.8. For 3.8 I am not sure about the value of porting this change, as it will be EOL by December of this year (and very unlikely that we do any release of the tool before that)
> 
> That's unfortunate... My intention was to push the 3.9.0 build to add arm support to the current [generate-delta-worker pipeline](https://github.com/mendersoftware/generate-delta-worker/blob/master/.gitlab-ci.yml#L3) which uses the `3.9.0`. So now I guess we have to move the generate-delta-worker to `3.10.1`?

I see. This is out of the release process as per today (tagged client components get published from mender-qa, never from their individual pipelines) so it wouldn't have worked anyway but just adding this here. I never really understood what was the motivation.

As the process is today, such a multi platform build (for client components, at least) needs to be done in mender-qa pipeline. Same for `mender-cli`, for example.

I suggest we bring this topic tomorrow in Mender QA meeting and we can take an educated decision with Client team and SRE team inputs on the matter :smiley:  I'll add a point in the agenda.

PR #547, opened 2023-09-22
3.10.x multiplatform build

oldgiova:
Ticket: QA-613

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -25,7 +25,6 @@ cache:
     - /go/src/gopkg.in
 
 variables:
-  DOCKER_HOST: tcp://docker:2375/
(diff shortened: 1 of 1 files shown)


lluiscampos:
@oldgiova Thanks. Please use the `-x` option for `git commit` when cherry-picking. This is the best way to track that this has already been approved. When approved in master and the commit applies clean in the release branch, we don't need re-approval :)

Also, what we usually do however is to trigger the cherry-picks using the bot (like in [here](https://github.com/mendersoftware/mender-artifact/pull/542#issuecomment-1732236689)) so that the author can approve and merge himself.

lluiscampos:
Closing in favor of https://github.com/mendersoftware/mender-artifact/pull/548

PR #561, opened 2023-11-27
MEN-5463 use source file name if destination file name is not supplied

NivKeidan:
https://northerntech.atlassian.net/browse/MEN-5463

### DIFF
--- a/cli/copy.go
+++ b/cli/copy.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


cli/copy.go
```diff
@@ -94,7 +95,14 @@ func Copy(c *cli.Context) (err error) {
 		if err != nil {
 			return cli.NewExitError(err, 1)
 		}
-		vfile, err = virtualImage.OpenFile(privateKey, c.Args().Get(1))
+
+		// if destination file name is not supplied, use source file name MEN-5463
```
lluiscampos, line 99: The comment is nice to have (thanks!), but the reference to the bug tracker is not needed
```suggestion
		// if destination file name is not supplied, use source file name
```

lluiscampos (approved):

lluiscampos:

lluiscampos:
@NivKeidan thanks again for your contribution! It is merged now and it will be part of the next release.

PR #563, opened 2023-12-04
QA-639: Explicitly set required `images` per job

lluiscampos:
The previous usage of a default `image` was correct for this repository, but has unfortunately been ported to other repositories where it miss-behaved.

Clean-up a bit the images used by CI with the following ideas:
* For the unit tests, explicitly set the golang version to use
* For non-build jobs, use base Debian images instead of golang ones
* Where possible, use `-slim` versions of the images

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,9 +15,6 @@ include:
   - project: "Northern.tech/Mender/mendertesting"
     file: ".gitlab-ci-check-license.yml"
 
-# Keep golang version aligned with latest yocto release
-image: golang:1.17-alpine3.16
-
 cache:
   paths:
     - /go/src/github.com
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -194,6 +193,8 @@ test:unit:mac:
 
 publish:tests:
   stage: publish
+  # Keep golang version aligned with latest yocto release
+  image: golang:1.20-alpine3.17
```
oleorhagen, line 197: Q: So not kirkstone? I'm pretty sure that is 1.17

oleorhagen (approved):

.gitlab-ci.yml
lluiscampos: Thanks. It is the comment that is off here, I meant to add such comment to `test:smoketests:linux:`. For this job, it is okey to use newer golang (and it is convenient to use newer to get `goveralls` working properly).

See rework.

oleorhagen: Makes sense now :)

PR #572, opened 2023-12-15
feat: adding functionality for adding script using script flag while …

SebOpsahl:
…in modify

changelog: title
ticket: MEN-5967


README.md
```diff
@@ -144,7 +144,6 @@ Auto-completion for ZSH is supported through the `zsh_autocompletion` script
 found in the `./autocomplete` directory. In order to enable it consistently, add
 these lines to your `.zshrc` file:
 
-```bash
```
kacf, line 147: Was this a mistake? 🙂

artifact/scripter.go
```diff
@@ -43,7 +43,6 @@ func (s *Scripts) Add(path string) error {
 	if s.names == nil {
 		s.names = make(map[string]string)
 	}
-
```
kacf, line 46: There are some whitespace changes here that don't really belong here.

It's ok to do cleanups if you see something which is worth changing, but then they should be in their own commit.

artifact/scripter.go
```diff
@@ -79,4 +78,4 @@ func (s *Scripts) Get() []string {
 		scr = append(scr, script)
 	}
 	return scr
-}
+}
```
kacf, line 81: Same here, unnecessary whitespace changes.

cli/artifacts.go
```diff
@@ -54,7 +54,6 @@ type writeUpdateStorer struct {
 func (w *writeUpdateStorer) Initialize(artifactHeaders,
 	artifactAugmentedHeaders artifact.HeaderInfoer,
 	payloadHeaders handlers.ArtifactUpdateHeaders) error {
-
```
kacf, line 57: Same here, unnecessary whitespace changes.

cli/artifacts.go
```diff
@@ -346,7 +345,6 @@ func reconstructPayloadWriteData(
 			return
 		}
 	}
-
```
kacf, line 349: Same here, unnecessary whitespace changes.

cli/artifacts_test.go
```diff
@@ -70,7 +69,6 @@ func WriteTestArtifact(version int, update string, key []byte) (io.Reader, error
 			return nil, errors.Wrap(err, "artifact.NewPKISigner")
 		}
 		aw = awriter.NewWriterSigned(buff, comp, s)
-		fmt.Println("write signed artifact")
```
kacf, line 73: This should probably stay, or?

cli/cli.go
```diff
@@ -252,11 +257,6 @@ func getCliContext() *cli.App {
 		privateKeyFlag,
 		gcpKMSKeyFlag,
 		vaultTransitKeyFlag,
-		cli.StringSliceFlag{
-			Name: "script, s",
-			Usage: "Full path to the state script(s). You can specify multiple " +
-				"scripts providing this parameter multiple times.",
-		},
```
kacf, line 259: It should also stay here, shouldn't it?

It's tempting to use the `artifactAddScripts` shared variable, but the help text is not appropriate for this location, since it is for new artifacts. So maybe it's best to avoid that and restore the old code.

cli/modify.go
```diff
@@ -254,6 +254,15 @@ func modifyArtifactAttributes(c *cli.Context, image VPImage) error {
 		art.writeArgs.Depends.ArtifactGroup = c.StringSlice("depends-groups")
 	}
 
+	if c.IsSet("script") {
+		if !isArt {
```
kacf, line 258: Good catch, I had already forgotten about this!

cli/modify.go
```diff
@@ -320,7 +329,7 @@ func modifyPayloadProvidesDepends(c *cli.Context, image VPImage) error {
 		}
 		art.writeArgs.AugmentTypeInfoV3.ArtifactDepends = typeInfoDepends
 	}
-
+	//double
```
kacf, line 332: Remove?

cli/modify.go
```diff
@@ -390,4 +399,4 @@ func modifyPayloadClearsProvides(c *cli.Context, image VPImage) error {
 	}
 
 	return nil
-}
+}
```
kacf, line 402: Unnecessary whitespace change.

cli/modify_existing_test.go
```diff
@@ -497,6 +496,7 @@ func TestModifyModuleArtifact(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Modify Artifact name shall work
+	//find structure and where to add it, in action it would be '/tmp/mender-artifact1689804532/scripts/'
```
kacf, line 499: Not sure if I understand this comment. Was it for your own consumption?

cli/partition.go
```diff
@@ -458,7 +458,6 @@ func filterSparsePartitions(parts []partition) []partition {
 // for rootfs{a,b}, this is the two partitions (unless one of them is unpopulated,
 // then only the one with data is returned)
 func getFilesystems(fpath string, modcands []partition) ([]partition, string) {
-
```
kacf, line 461: Unnecessary whitespace change.

vendor/github.com/urfave/cli/flag_string.go
```diff
@@ -76,6 +76,10 @@ func (c *Context) String(name string) string {
 	return lookupString(name, c.flagSet)
 }
 
+func (c *Context) StringMultiple(name string) string {
```
kacf, line 79: Files in vendor folders are dependencies that we fetch from elsewhere, and periodically sync. Therefore the files are just copies, and we cannot change them, because it would be overwritten on the next update. Therefore changes to vendor folders are not allowed.

If a change in a vendor folder is needed, it needs to go through that library, so we need to submit a pull request to that repository instead, and update to it after it has been accepted.

It looks like you are not using this function though, nor any other change that you made in the vendor folder, so you can just revert all those changes to their original state.

kacf (changes requested):
Nice work! The main code is exactly as it should be, small and tidy! But there are some small adjustments needed, which you'll see in the comments below.

I'm requesting a lot of whitespace cleanup. One useful trick I can give you is to execute:

```
git checkout -p HEAD~1
```

What this means is: Take the commit before this one (that's `HEAD~1`), and ask one-by-one if you want to restore the content from that commit. Then you can selectively choose:

* "y" - Yes, please revert this
* "n" - No, I want to keep this

Another useful trick, after you've made all the changes, is to run `git add -p`. This again asks you, one-by-one, which changes you want to add to the commit. When you're done with it, use `git commit --amend` and `git push` again to update the pull request.

text.txt
kacf, line 1: You didn't mean to submit this one, did you? 🙂

PR #579, opened 2024-01-03
feat: adding functionality for adding script using script flag while …

SebOpsahl:
…in modify

changelog: title
ticket: MEN-5967

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -162,6 +162,7 @@ func getCliContext() *cli.App {
 		Usage:    "Name of the artifact",
 		Required: true,
 	}
+
 	artifactNameDepends := cli.StringSliceFlag{
 		Name:  "artifact-name-depends, N",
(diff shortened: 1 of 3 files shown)


kacf (changes requested):
Hey @SebOpsahl, something has happened with the commit structure, and this is now a merge commit, which is not appropriate. The author has also gotten messed up (says Lluis).

In general we don't use merges when pulling new changes before submitting to upstream. Instead we use rebasing. It might be useful to set this:

```bash
git config --global pull.rebase true
```

which will cause `git pull` to default to rebasing instead of merging.

As for fixing this pull request, I propose the following steps:

```bash
# Resets history back to the upstream original,
# but keeps the changes you made.
git reset --soft origin/master

# Make a new commit, with fresh properties
git commit -s

# Update PR
git push -f SebOpsahl HEAD:new-feat-modify-script-mender-artifact
```

kacf (approved):
Great!

SebOpsahl:

PR #598, opened 2024-03-14
chore: omit mender-snapshot path from output

danielskinstad:
The path to mender-snapshot was printed when running 'which mender-snapshot'. Now redirecting stdout to /dev/null when checking the path.

Ticket: MEN-7107


lluiscampos (approved):
Thank you. So simple :+1:

danielskinstad:

PR #604, opened 2024-04-25
Add build constraint to allow compiling without `cgo`

alfrunes:
The PKCS11 feature (not used by the backend) prevents compiling mender-artifact without enabling `cgo`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Keep golang version aligned with latest yocto release
-FROM golang:1.17.13-bullseye as builder
+FROM golang:1.22.2-bullseye as builder
 RUN mkdir -p /go/src/github.com/mendersoftware/mender-artifact
 WORKDIR /go/src/github.com/mendersoftware/mender-artifact
 ADD ./ .
--- a/Dockerfile.binaries
+++ b/Dockerfile.binaries
@@ -1,5 +1,5 @@
 # Keep golang version aligned with latest yocto release
-FROM golang:1.17.13-bullseye as builder
+FROM golang:1.22.2-bullseye as builder
 RUN apt-get update && \
     apt-get install -y \
     gcc gcc-mingw-w64 gcc-multilib \
--- a/artifact/signer_linux.go
+++ b/artifact/signer_linux.go
@@ -12,8 +12,8 @@
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
 
-//go:build linux
-// +build linux
+//go:build linux && cgo
+// +build linux,cgo
 
 package artifact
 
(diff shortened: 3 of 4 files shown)


tranchitella (approved):

kacf (changes requested):
I think this should be made into a feature flag (such as `nopkcs11` or something like that). I think silently disabling features if cgo is disabled is surprising and unintuitive.

However, independent from that, I don't think it will work as intended, because I think lzma requires cgo as well, and we can't disable that. Not 100% sure though.

alfrunes:
> I think this should be made into a feature flag (such as `nopkcs11` or something like that). I think silently disabling features if cgo is disabled is surprising and unintuitive.

This is exactly how we do it for `lzma` (I found out after filing this PR), and I agree that disabling features should be explicit. However, this feature abuse the built-in `linux` constraint which makes impossible to opt out (in the backend we do not use this dependency). I will change the build tag to `pkcs11` and have it set by default in the Makefile.

> However, independent from that, I don't think it will work as intended, because I think lzma requires cgo as well, and we can't disable that. Not 100% sure though.

For `lzma` I will similarly add a different (Go) dependency if compiled without cgo since this one is crucial.

alfrunes:
Thoughts on this one? #605

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Keep golang version aligned with latest yocto release
-FROM golang:1.17.13-bullseye as builder
+FROM golang:1.22.2-bullseye as builder
```
lluiscampos, line 2: Please keep the version aligned to yocto, see comment above.

Coincidentally, we are discussing this in the [upcoming client team meeting](https://docs.google.com/document/d/1neXnS16brsb7Jdn-Sg-f1Wyoyo_O5MJIAW7_FzTKpfY/edit#heading=h.ih2nzgp6q407).

lluiscampos (changes requested):

Dockerfile
alfrunes: Please keep in mind that `go1.17` has been out of support for almost 2 years. Is it not possible to use the most recent version of Go when compiling Yocto?

lluiscampos: Yocto maintains the toolchain, porting security fixes (same as for example Debian or other OSs do for the same time period that they support the OS).

Fixing the version here does have any impact in the final yocto build nor in the deb packages build.

Fixing the version here is only a way for us (me) to early catch when updates will break in today *latest* yocto LTS. Then we can decide to accept the breakage and patch the code there or reject the update for the time being. It is a difficult balance - to give you a small nitpick: we still support `dunfell` LTS which uses golang 1.14 and it is really hard and time consuming delivering our software there.

alfrunes:
Superseded by #605

PR #605, opened 2024-04-25
Allow using mender-artifact without `cgo`

alfrunes:
* Adds an alternative dependency for `lzma` with `cgo` disabled
* Disables `pkcs11` if built with `nopkcs11` build tag.


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Keep golang version aligned with latest yocto release
-FROM golang:1.17.13-bullseye as builder
+FROM golang:1.22.2-bullseye as builder
```
lluiscampos, line 2: Same as in [the previous PR](https://github.com/mendersoftware/mender-artifact/pull/604/files#r1580639114) :smiley:

artifact/compressor_lzma.go
```diff
@@ -11,15 +11,16 @@
 //    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 //    See the License for the specific language governing permissions and
 //    limitations under the License.
-//go:build !nolzma && cgo
-// +build !nolzma,cgo
+//go:build !nolzma && !cgo
+// +build !nolzma,!cgo
```
lluiscampos, line 15: I am actually in favor of replacing LZMA compression from shared to static linked version (this one, and drop the old code) instead of having the two with compile flags.

Pros:
* Safer testing (as it stands now, for example, the acceptance and integration tests will only test the cgo version)
* One less OS dependency (`xz` is a pretty safe dependency, but still)

Cons:
* Bigger binary size (but it can still be opted out with `nolzma` flag)

kacf: Agree, let's just ditch the old one. Size is even less important now because the C++ client does not use mender-artifact to parse. Then we can remove `cgo` completely from the equation here.

Makefile
```diff
@@ -28,7 +28,7 @@ ifeq ($(V),1)
 BUILDV = -v
 endif
 
-TAGS =
+TAGS ?= pkcs11
```
kacf, line 34: 👍

kacf (approved):
Pre-approving, but see comments below.

Makefile
alfrunes: Thank you for pointing the thumb at a left-over :sweat_smile: 
I actually did the inverse build constraints first before I changed the tag to `nopkcs11` :upside_down_face:

kacf: Oh, I thought you did it for documentation purposes! But yes, maybe you're right, it's better to remove it! 😄

artifact/compressor_lzma.go
alfrunes: Done

alfrunes:

Makefile
```diff
@@ -144,15 +146,7 @@ instrument-binary:
...
-	echo 'mode: count' > coverage.txt
-	set -e ; for p in $(PKGS); do \
-		rm -f coverage-tmp.txt;  \
-		$(GO) test -covermode=count -coverprofile=coverage-tmp.txt -coverpkg=$(SUBPKGS) $$p ; \
-		if [ -f coverage-tmp.txt ]; then \
-			cat coverage-tmp.txt |grep -v 'mode:' | cat >> coverage.txt; \
-		fi; \
-	done
-	rm -f coverage-tmp.txt
+	go test -tags '$(TAGS)' -covermode=atomic -coverprofile=coverage.txt ./...
```
lluiscampos, line 149: I think that this is the reason for the coverage drop. I don't remember the details, but I looked in this years ago and ended up concluding that it is necessary to do package by package :man_shrugging:

alfrunes:

alfrunes:

alfrunes:
Turns out that the BinaryTree matcher for the new lzma implementation (github.com/ulikunitz/xz) seems to be hanging when compressing objects. I changed the default config to use HashTable4 (hc4 in xz(1)) instead.

lluiscampos (approved):

PR #610, opened 2024-05-07
ci: Move all jobs to private runners

lluiscampos:


oldgiova (approved):

oldgiova (approved):

kacf (approved):

lluiscampos:
The jobs are running into other sorts of errors. I'll revert back to just updating the public shared runners tag and wait for a final decision on moving or not to private runners.

lluiscampos:
> (...) I'll revert back to just updating the public shared runners tag (...)

See https://github.com/mendersoftware/mender-artifact/pull/611

PR #613, opened 2024-05-27
chore(Dockerfile): Refactor dockerfile to Debian Slim with no deps

alfrunes:


Dockerfile
```diff
@@ -1,9 +1,10 @@
 # Keep golang version aligned with latest yocto release
```
lluiscampos, line 1: We discussed this in [the past](https://github.com/mendersoftware/mender-artifact/pull/604/files#r1580639114). We cannot silently update the golang version.

However the intention now is to actually publish this image, so we cannot anymore lock to golang 1.17 here (obviously). So could you please extend your PR (scope creep coming!) with the following:
* Add a new Dockerfile (f.eks `Dockerfile.backcompat` or something like that) which still build the binary with golang 1.17 and keeps this comment
* Add a new CI job to just do `docker build` on this new Dockerfile

lluiscampos (changes requested):
Okey, so we are finally going to publish `mender-artifact` on Docker too :+1: 

See below - we need to preserver the backwards compatibility build before merging this.

Dockerfile
lluiscampos: > Having an extra Dockerfile for building for go1.17 seems a bit excessive. How about a simple gitlab job (`build:yoctocompat`) that runs `go build` in a `golang:1.17` container?

Actually I thought about that during this week of inactivity in this PR. It is not only better, but also aligned to other client repos when we had to do backwards compatibility jobs. Thanks

.gitlab-ci.yml
```diff
@@ -28,27 +28,41 @@ variables:
...
 build:docker:
-  image: docker
+  image: docker:26-cli
   needs: []
   services:
-    - docker:dind
+    - docker:26-dind
   stage: build
+  before_script:
+    - docker login --username "$CI_REGISTRY_USER" --password "$CI_REGISTRY_PASSWORD" "$CI_REGISTRY"
```
lluiscampos, line 46: If the password is not masked, this line would leak it. It is better to use

```suggestion
    - echo ${CI_REGISTRY_PASSWORD} | docker login --username ${CI_REGISTRY_USER} --password-stdin ${CI_REGISTRY}
```

.gitlab-ci.yml
```diff
@@ -260,6 +274,30 @@ publish:s3:
...
+    name: "quay.io/skopeo/stable:v1.15"
+    # https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#override-the-entrypoint-of-an-image
+    entrypoint: [""]
+  before_script:
+    - skopeo login --username "$CI_REGISTRY_USER" --password "$CI_REGISTRY_PASSWORD" "$CI_REGISTRY"
+    - |
+      skopeo login \
+        --username "$DOCKER_HUB_USERNAME" \
+        --password "$DOCKER_HUB_PASSWORD" \
+        docker.io
```
lluiscampos, line 294: I don't know if `skopeo login` supports also `--password-stdin`, but if so please use it here too

Dockerfile
```diff
@@ -1,9 +1,11 @@
...
```
lluiscampos, line 1: Remove this comment from here :smiley:

.gitlab-ci.yml
```diff
@@ -28,27 +28,41 @@ variables:
...
+    - |
+      docker buildx build --push \
+        --platform "$DOCKER_PLATFORMS" \
+        --tag "$CI_REGISTRY_IMAGE:$CI_PIPELINE_IID" \
+        .
   tags:
     - gitlab-org-docker
 
+build:yoctocompat:
+  image: golang:1.17.13-bullseye
```
lluiscampos, line 58: And add a comment here
```suggestion
  # Keep golang version aligned with latest yocto release
  image: golang:1.17.13-bullseye
```

lluiscampos (approved):
Thank you! I am very happy to see this coming up together :rocket: 

Pre-approving, but please see my comments below

.gitlab-ci.yml
alfrunes: I'm not sure I get it. Isn't it the same with the echo command? The variable is expanded there as well.

.gitlab-ci.yml
alfrunes: It does, the CLI is pretty much identical.

.gitlab-ci.yml
lluiscampos: It won't be logged into stdout - (iow  the CI logs). I cannot link to an LFF - but search for "GitLab CI secrets leak" in the LFF archive :)

lluiscampos (approved):

alfrunes:
@lluiscampos I'm wondering if we should bump the `build:make` job to `go1.23.4` since these binaries are not used inside Yocto. Otherwise, I'd remove the `build:yoctocompat` job since it would otherwise not be needed. What do you think?

lluiscampos:
> @lluiscampos I'm wondering if we should bump the `build:make` job to `go1.23.4` since these binaries are not used inside Yocto. Otherwise, I'd remove the `build:yoctocompat` job since it would otherwise not be needed. What do you think?

Yes, there is no reason to keep `build:make` in older golang.

As a rule of thumb, we should have use golang versions: one specific for the new `build:yoctocompat` job (go 1.17, although in a matter of weeks we will be able to upgrade!), and then the latest for every other job and Dockerfiles :+1:

alfrunes:
> @lluiscampos I'm wondering if we should bump the `build:make` job to `go1.23.4` since these binaries are not used inside Yocto. Otherwise, I'd remove the `build:yoctocompat` job since it would otherwise not be needed. What do you think?

Oops, sorry (apparently my future self was talking here). I meant `go1.22.4` of course :sweat_smile:

> Yes, there is no reason to keep `build:make` in older golang.
> 
> As a rule of thumb, we should have use golang versions: one specific for the new `build:yoctocompat` job (go 1.17, although in a matter of weeks we will be able to upgrade!), and then the latest for every other job and Dockerfiles 👍

Agreed! I'll update the `build:make` job as well :+1:

alfrunes:

.gitlab-ci.yml
```diff
@@ -150,8 +150,6 @@ test:smoketests:linux:
     - ./mender-artifact-linux read test-rfs.mender
     - ./mender-artifact-linux validate test-rfs.mender
     - ./tests/test_compressions/test_supported_compressions.sh
-    # QA-507: lock mender-artifact to OpenSSL 1.1
-    - ldd ./mender-artifact-linux | grep libssl.so.1.1
```
lluiscampos, line 139: We should not break compatibility this way. We need to plan it a bit. As discussed in person, we could offer two flavors in our [Downloads page](https://docs.mender.io/downloads#mender-artifact): one statically linked without pkcs11 and one dynamically linked for openssl3.

As this starts to be bigger than a nice to have PR, can we create a task for it? It involves the work here, plus having some sort of subdirs structure in the S3 bucket, and `mender-docs`.

lluiscampos (changes requested):

alfrunes:
Closing this one as it's not a priority at the moment.

PR #633, opened 2024-09-18
QA-765: Explicit CI compatibility check + related updates

lluiscampos:

PR #635, opened 2024-09-20
Revert "Merge pull request #633 from lluiscampos/QA-765-explicit-comp…"

lluiscampos:
DO NOT MERGE


lluiscampos:
The failure is still present so it must be related to the overnight OS upgrade

PR #636, opened 2024-09-30
fix: hardware-security returned signatures support for ECDSA256.

merlin-northern:
Changelog: Title
Ticket: MEN-7523


artifact/signer.go
```diff
@@ -158,16 +160,52 @@ func MarshalECDSASignature(r, s *big.Int) ([]byte, error) {
...
-	s = big.NewInt(0).SetBytes(sig[ecdsa256keySize:])
-	return r, s, nil
+	// in case of a key supplied via PKCS#11 URI, we have no control over what the signature is
+	// since it is designed to be actually verified via the same mechanism (PKCS#11 URI).
+	// We know here that it is ECDSA key, and judging form the size we can assume
+	// that it is ASN.1 encoded. If so, then it should be 72 bytes, the less or equal is here
+	// to support keys of size 71 bytes (strangely seen in the wild). In other words:
+	// if the signature has not been created with MarshalECDSASignature, then we assume
+	// it is to be decoded via ASN.1, with the protection on the signature length.
+	if len(sig) <= ecdsa256Asn1SizeBytes {
```
lluiscampos, line 182: Based on the comment above, and with the aim of giving the best possible error messages to users, this should be:
```suggestion
	if (len(sig) == ecdsa256Asn1SizeBytes) || (len(sig) == ecdsa256Asn1SizeBytes - 1) {
```

artifact/signer.go
```diff
@@ -158,16 +160,52 @@ func MarshalECDSASignature(r, s *big.Int) ([]byte, error) {
...
+			_, err = asn1.Unmarshal(bytesValue, &v)
+			if err != nil {
+				return nil, nil, errors.Wrap(err, "cannot unmarshal asn1 s value")
+			}
+			s = big.NewInt(0).SetBytes(v.Bytes)
+		}
+		return r, s, nil
+	}
+
+	return nil, nil, errors.Errorf("signer: invalid ecdsa key size: %d", len(sig))
```
lluiscampos, line 208: Please consider refactoring the whole thing.
```
func UnmarshalECDSASignature(sig []byte) (r, s *big.Int, e error) {
    if len(sig) == 2*ecdsa256keySize {
        return  UnmarshalECDSASignatureMenderArtifact(...)
    }
    if len(sig) <= ecdsa256Asn1SizeBytes {
        return  UnmarshalECDSASignatureASN1(...)
    }
    return error...
```

lluiscampos (commented):
It looks good. The comments are great :rocket:

artifact/signer.go
lluiscampos: Reasoning: a shorter signature should say to the user "invalid ecdsa key size" rather than "cannot unmarshal asn1 data"

artifact/signer.go
lluiscampos: Reasoning: it will make the code more readable and the comments contained inside separate functions

artifact/signer.go
merlin-northern: > Based on the comment above, and with the aim of giving the best possible error messages to users, this should be:

fixed.

merlin-northern:
@mender-test-bot start pipeline please with cherry on top

merlin-northern:
@lluiscampos heads up: I added smoke test with my public key from my nitro and an artifact signed with the private key from the same nitro :>

.gitignore
```diff
@@ -2,7 +2,6 @@
 *.o
 *.a
 *.so
-*.mender
```
lluiscampos, line 5: Unrelated change

ec.pem
```diff
@@ -0,0 +1,4 @@
+-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEDTjFpi9+tJ8KnTPQff34Gpj4E4uc
+T4ML17IzkcWZhtY1O44fqMHmejS0umai8xJlD0Z/THhTiuRO8PC3vpkU9g==
+-----END PUBLIC KEY-----
```
lluiscampos, line 4: This one and `a0-signed-nitro.mender` files should be in a subdirectory inside `tests/`, ideally with an small README on how they have been generated.

artifact/signer.go
```diff
@@ -158,11 +160,56 @@ func MarshalECDSASignature(r, s *big.Int) ([]byte, error) {
...
 
+	// in case of a key supplied via PKCS#11 URI, we have no control over what the signature is
+	// since it is designed to be actually verified via the same mechanism (PKCS#11 URI).
+	// We know here that it is ECDSA key, and judging form the size we can assume
+	// that it is ASN.1 encoded. If so, then it should be 72 bytes, the less or equal is here
+	// to support keys of size 71 bytes (strangely seen in the wild). In other words:
+	// if the signature has not been created with MarshalECDSASignature, then we assume
+	// it is to be decoded via ASN.1, with the protection on the signature length.
+	if len(sig) <= ecdsa256Asn1SizeBytes ||
+		len(sig) <= ecdsa256Asn1SizeBytes-1 {
```
lluiscampos, line 179:
```suggestion
	if len(sig) == ecdsa256Asn1SizeBytes ||
		len(sig) == ecdsa256Asn1SizeBytes-1 {
```

lluiscampos (changes requested):
This fix deserves a better changelog entry :pray: 

For example: "Fixes signature verification with ECDSA keys when the signing has been done externally with PKCS#11. Previously `mender-artifact` would always assume that the signature has been done with the built-in engine, which then wouldn't validate correctly. The bug affected only ECDSA key pairs."

artifact/signer.go
merlin-northern: wooops sorry, copy-paste.

ec.pem
merlin-northern: none of the other artifacts in the smoke tests are there. one day we must make order here. what I did last time I was here was to introduce the tests folder and the scripting around the tests. I moved the key and the artifact to the test/data directory and I added a read me.

.gitignore
merlin-northern: without it I will not be able to add the artifacts.

lluiscampos: Yes, but it deserves a separate commit or a mention in the commit description.

ec.pem
lluiscampos: > none of the other artifacts in the smoke tests are there. 

What do you mean? Which other Artifacts? As far as I can see this is the only checked in artifact in the repo :mag_right: 

> (...), I moved the key and the artifact to the test/data directory and I added a read me

Thanks

lluiscampos (approved):

ec.pem
merlin-northern: artifacts: ignore me it was a complicated situation, you are right. and it would take me so long to properly explain myself. my mistake in general.

.gitignore
merlin-northern: separate commit added

merlin-northern:
I am not running the full pipeline as it already passed with the same code.

merlin-northern:
 * 3.11.x

please

merlin-northern:
> Hi 😺 I did my very best, and this is the result of the cherry pick operation:
> 
> * 3.11.x Had merge conflicts, you will have to fix this yourself 😿

https://github.com/mendersoftware/mender-artifact/pull/639

merlin-northern:
> @merlin-northern Is this fix in the version from master that is linked in the documentation? Is there some way I would know when that linked master version is updated?
> 
> https://docs.mender.io/downloads#mender-artifact

you need someone to verify -- both master and the 3.11.x -- I will not issue an official message without double checking. I suppose I can do it, and for sure Client Team can do it. so, take your pick.

lluiscampos:
> @merlin-northern Is this fix in the version from master that is linked in the documentation? Is there some way I would know when that linked master version is updated?
> 
> https://docs.mender.io/downloads#mender-artifact

Yes it is. As well as in the APT repositories experimental distributions (which are updated once a day)

PR #645, opened 2024-10-16
Use golang's crypto packages instead of mendersoftware/openssl

siredmar:
This PR replaces the dependency for the ancient [github.com/mendersoftware/openssl](https://github.com/mendersoftware/openssl) package. Why?

Because this openssl library requries CGO_ENABLED which can be very tricky on embedded systems. 
Using golang's std crypto packages there is no need for CG_ENABLED and the binary can easily cross compiled for different environments. 

There are no changes from user perspective: signing and validating works just the same using the same privat/public keys.


siredmar:
nevermind... there seems to be no go native pcks#11 implementation existing.

lluiscampos:
Yes, exactly!

We also would like to move away from openssl but as you found out the issue we are trying to cover with openssl is support for hardware security modules via PKCS#11

PR #646, opened 2024-10-24
Do not replace output file on write

alfrunes:
The `-o` / `--output-path` argument creates a new file (`.tmp`) and replaces the target path with the newly created file. This behavior is unexpected especially if you want to pipe the output to another process.
I changed the behavior to instead write to the destination directly (even if it fails half way).


lluiscampos (changes requested):
`mender-artifact` 4.0.0 coming up :rocket: 

I am all :+1: for this change (thanks!), but giving the public impact of it (users and CE will ask why a major bump) we need:

* A JIRA task for it.
* A more elaborated changelog entry for the breaking change
    * Suggestion: "`mender-artifact` write commands no longer write to a temporary file and then rename to the user specified file. While the user observerd behavior is the same on success, internally the tool is writing directly into the user specified file which means that will be left unusable on errors. This change is made in order to enable the tool for more advance uses like redirecting the output to a different process"

alfrunes:
> `mender-artifact` 4.0.0 coming up 🚀
> 
> I am all 👍 for this change (thanks!), but giving the public impact of it (users and CE will ask why a major bump) we need:
> 
> * A JIRA task for it.

Will do. I just wanted to get some feedback first as I didn't really know if there's a reason why we did this in the past.

> * A more elaborated changelog entry for the breaking change
>   
>   * Suggestion: "`mender-artifact` write commands no longer write to a temporary file and then rename to the user specified file. While the user observerd behavior is the same on success, internally the tool is writing directly into the user specified file which means that will be left unusable on errors. This change is made in order to enable the tool for more advance uses like redirecting the output to a different process"

I wasn't sure about putting the "BREAKING CHANGE" trailer for this change. Although it is a change in behavior, the expected behavior from the API is to overwrite the destination file anyway - so from a different angle it's fixing the behavior that it's replacing the destination file and instead write to it.

alfrunes:
I created [MEN-7660](https://northerntech.atlassian.net/browse/MEN-7660) and [MEN-7661](https://northerntech.atlassian.net/browse/MEN-7661) and removed the `BREAKING CHNAGE` annotation.

alfrunes:

lluiscampos:
> I wasn't sure about putting the "BREAKING CHANGE" trailer for this change. Although it is a change in behavior, the expected behavior from the API is to overwrite the destination file anyway - so from a different angle it's fixing the behavior that it's replacing the destination file and instead write to it.

Yes, I also saw that dilemma. It depends on the focus: from a user in the success case there is no change (that is what I wanted to highlight on my suggested changelog entry) but technically speaking there is a breaking change that can be unexpected.

For example: I create a valid Artifact, I get the `artifact.mender` file. I attempt to create another, now it fails. The previous `artifact.mender` is overridden with an incomplete Artifact, which was not the case with previous versions of the tool.

All in all I thin we should be in the safe side and mark it as breaking change.

lluiscampos (approved):
Sorry, I don't know how... but this PR felt out of my radar! :radio: I was so sure that I commented before your last rework (??).

Anyway, I still think it is a breaking change, but I won't push farther. The client acceptance tests have been in trouble for other reasons, let me run the pipeline for you and merge :+1: Thanks again for doing this work.

PR #659, opened 2024-12-30
fix: fixed signing artifact via symlink

bahaa-ghazal:
Changelog: Title
Ticket: MEN-3410

### DIFF
--- a/cli/sign.go
+++ b/cli/sign.go
@@ -60,9 +60,28 @@ func signExisting(c *cli.Context) error {
 	}
 	defer f.Close()
 
-	artFileStat, err := os.Stat(artFile)
+	artFileStat, err := os.Lstat(artFile)
 	if err != nil {
 		return cli.NewExitError("Could not get artifact file stat", 1)
+	}
+
+	if artFileStat.Mode()&os.ModeSymlink == os.ModeSymlink {
+		f.Close()
+		artFile, err = os.Readlink(artFile)
+		if err != nil {
+			return cli.NewExitError(err, 1)
+		}
+		outputFile = artFile
+		artFileStat, err = os.Stat(artFile)
+		if err != nil {
+			return cli.NewExitError("Could not get artifact file stat", 1)
+		}
+		f, err = os.Open(artFile)
+		if err != nil {
+			err = errors.Wrapf(err, "Can not open: %s", artFile)
+			return cli.NewExitError(err, 1)
+		}
+		defer f.Close()
 	}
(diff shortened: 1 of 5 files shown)


lluiscampos (commented):
Can you add a test? :pray:

lluiscampos (approved):
Thank you!

There are some static issues that you need to fix, but the code looks good :rocket:

cli/sign.go
```diff
@@ -58,12 +58,30 @@ func signExisting(c *cli.Context) error {
 		err = errors.Wrapf(err, "Can not open: %s", artFile)
 		return cli.NewExitError(err, 1)
 	}
-	defer f.Close()
```
danielskinstad, line 61: I think keeping this defer here and adding a new `defer f.Close()` right after the new `Open` as it's done here is a bit more clear

cli/sign.go
```diff
@@ -58,12 +58,30 @@ func signExisting(c *cli.Context) error {
...
 
-	artFileStat, err := os.Stat(artFile)
+	artFileStat, err := os.Lstat(artFile)
 	if err != nil {
 		return cli.NewExitError("Could not get artifact file stat", 1)
```
danielskinstad, line 64: The file is not closed here, but keeping the old defer right after would solve it

cli/sign.go
```diff
@@ -58,12 +58,30 @@ func signExisting(c *cli.Context) error {
...
+		outputFile = artFile
+		artFileStat, err = os.Stat(artFile)
+		if err != nil {
+			return cli.NewExitError("Could not get artifact file stat", 1)
+		}
+		f, err = os.Open(artFile)
+		if err != nil {
+			err = errors.Wrapf(err, "Can not open: %s", artFile)
+			return cli.NewExitError(err, 1)
+		}
```
danielskinstad, line 82: I would suggest having a second defer here, and remove the one right below this, as I mentioned above

danielskinstad (changes requested):
Some comments about the newest commit :), you can see [this](https://github.com/mendersoftware/mendertesting/blob/master/commitlint/grammar.md) for when to use what
* The error formatting fixes some static checks, right? Fixes are generally reserved for bugs in the code
* I don't think the format refactoring commit warrants a changelog entry, so I would set it to `None` instead of `Title`.  
* Typo `erorrs` instead of `errors` 

And a small comment/suggestion regarding closing, otherwise I think it looks good :smile:

PR #662, opened 2025-01-09
chore: Improving error handling when `blkid` is missing

bahaa-ghazal:
Changelog: Title
Ticket: MEN-4977


bahaa-ghazal:

cli/debugfs_test.go
```diff
@@ -62,5 +62,5 @@ func TestExternalBinaryDependency(t *testing.T) {
 	assert.Contains(t, err.Error(), "`parted` binary not found on the system")
```
lluiscampos, line 62: Once we rename "on the system" to "on this machine" for errors on not found binaries, we should do it consistently for all binaries.

Iow, here `parted` should also be updated and we should look for more instances in the code base.

(if you take this suggestion, please do it as a separate commit)

lluiscampos (changes requested):
Thanks you for tackling this one.

First of all, this PR does not resolve the issue reported in MEN-4977. If you pay attention to the description there, the command exists after "*error:* `blkid` binary not found", while your PR addresses the case where the *warniing* "Skipping running fsck..." is printed after the snapshot is done.

Now, I don't know how such an error is triggered. On my machine I removed `blkid` and run the command and it correctly prints the warning instead of the error :thinking: Maybe there is some another error really happening, but it is incorrectly reported by `imgFileSystemType` ?

Secondly, as a general comment, please let's strive for more content rich commit messages: "Improving error handling" is subjective to me - improving how? The commit title can be "improving something", fair enough, but then in the the body you could write "by early printing the warning when the binary tool is not found, and renaming confusing "system" to more explicit "this machine". Or something like that. At least a user should be able to relate what "improvements" this code change is bringing.

danielskinstad:
> Now, I don't know how such an error is triggered. On my machine I removed `blkid` and run the command and it correctly prints the warning instead of the error 🤔 Maybe there is some another error really happening, but it is incorrectly reported by `imgFileSystemType` ?

It logs the warning instead, because the function was changed a month after the ticket was created; https://github.com/mendersoftware/mender-artifact/commit/15d666f2b9c578bfb98f898189ef0590c678f708

lluiscampos (approved):
> It logs the warning instead because the function was changed a month after the ticket was created; [15d666f](https://github.com/mendersoftware/mender-artifact/commit/15d666f2b9c578bfb98f898189ef0590c678f708)

Great, thanks.  :see_no_evil: 

It was the JIRA ticket that confused me - we could have commented to the author there then that the error part of issue was actually solved already.

@bahaa-ghazal My comment wrt the commit content, plus the comment wrt consistency in the naming system/machine still stand, but they are minor suggestions so pre-approving here. Good work

cli/write.go
```diff
@@ -1005,6 +1007,7 @@ func recvSnapshot(dst io.Writer, src io.Reader) (int64, error) {
 	for {
 		nr, err := src.Read(buf)
 		if err == io.EOF {
+			fmt.Println()
```
danielskinstad, line 1010: I am guessing this is a left-over from something?

bahaa-ghazal: This is added because when the progress bar is finished, it doesn't create a newline at the end. This makes it so things are printed under the progress bar and not next to it.

bahaa-ghazal:
I was conflicted about doing this ticket because the error in the ticket was outdated where `blkid` wasn't needed to finshed process. I saw that the error was exchanged with a warning and got suggested moving it to before the SSH connection to make it less misleading. 

@lluiscampos, you're right. The commit message is too abstract and doesn't describe the changes well. I guess I was too focused on making it "short and sweet".  I'll rewrite the message asap and try to be more detailed 👍

bahaa-ghazal:

danielskinstad:
Remember to rebase your pr on top of master

bahaa-ghazal:

lluiscampos:
> I guess I was too focused on making it "short and sweet"

@bahaa-ghazal  There is always a compromise.

"Short and sweet" as you call it is good for today, while we humans look at this and we know the context. But in 5 years from now, someone else with no context will do a git blame on the code, find your commit, and wonder why did we do this change... So for that other scenario, "the more sugar the sweater" I would say :laughing:

PR #665, opened 2025-01-20
fix: Include all provides with prefix rootfs when writing artifact via SSH

bahaa-ghazal:
Changelog: Title
Ticket: MEN-7225

### DIFF
--- a/cli/write.go
+++ b/cli/write.go
@@ -268,12 +268,17 @@ func writeRootfs(c *cli.Context) error {
 		name = c.String("output-path")
 	}
 	version := c.Int("version")
+	var showprovides map[string]string
 
 	Log.Debugf("creating artifact [%s], version: %d", name, version)
 	rootfsFilename := c.String("file")
 	if strings.HasPrefix(rootfsFilename, "ssh://") {
 		rootfsFilename, err = createRootfsFromSSH(c)
 		defer os.Remove(rootfsFilename)
+		if err != nil {
+			return cli.NewExitError(err.Error(), errArtifactCreate)
+		}
+		showprovides, err = showProvides(c)
 		if err != nil {
 			return cli.NewExitError(err.Error(), errArtifactCreate)
 		}
@@ -335,6 +340,12 @@ func writeRootfs(c *cli.Context) error {
 	typeInfoV3, _, err := makeTypeInfo(c)
 	if err != nil {
 		return err
+	}
+	for k, v := range showprovides {
+		_, exist := typeInfoV3.ArtifactProvides[k]
+		if !exist {
+			typeInfoV3.ArtifactProvides[k] = v
+		}
 	}
 
 	if !c.Bool("no-checksum-provide") {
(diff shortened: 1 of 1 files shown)


bahaa-ghazal:

bahaa-ghazal:
There is a bug where the artifact name (flag -n) will be set to  the provide `rootfs-image.version`, which causes the `rootfs-image.version` to always be the artifact name (except if you give the `rootfs-image.version` with the `-p` flag ) and override the original `rootfs-image.version` from the `show-provide`. The occurs because the provides form `show-provides` should not override  any provides.
This bug could create inconsistency between the provides:
```
Mender Artifact:
  Name: foo
  Format: mender
  Version: 3
  Signature: no signature
  Compatible devices: [raspberrypi4]
  Provides group: 
  Depends on one of artifact(s): []
  Depends on one of group(s): []
  State scripts: []

Updates:
  - Type: rootfs-image
    Provides:
      rootfs-image.checksum: 5cfd893222cdf5a6503bd1f0faa90a79d90325250399a21e156a00ff2ebedd91
      rootfs-image.update-module.directory.mender_update_module: directory
      rootfs-image.update-module.directory.version: mender-image-5.0.0
      rootfs-image.update-module.mender-configure.mender_update_module: mender-configure
      rootfs-image.update-module.mender-configure.version: mender-image-5.0.0
      rootfs-image.update-module.rootfs-image.mender_update_module: rootfs-image
      rootfs-image.update-module.rootfs-image.version: mender-image-5.0.0
      rootfs-image.update-module.single-file.mender_update_module: single-file
      rootfs-image.update-module.single-file.version: mender-image-5.0.0
      rootfs-image.version: foo // should be 'mender-image-5.0.0'
    Depends: {}
    Clears Provides: [artifact_group, rootfs_image_checksum, rootfs-image.*]
    Metadata: {}
    Files:
      - checksum: 5cfd893222cdf5a6503bd1f0faa90a79d90325250399a21e156a00ff2ebedd91
        modified: 2025-01-20 13:00:02 +0100 CET
        name: rootfs.tmp4262774533
        size: 452984832

```

lluiscampos (commented):
The code looks good :100: 

I don't agree that it is a fix, it is rather a feature in my opinion. This is important because a fix we shall port it to stable branches, but in this cases I believe it is a feature and we leave it in `master` for the next minor release.

There is a drop in test coverage which is not desirable. We should have added a test in the [acceptance tests suite](https://github.com/mendersoftware/meta-mender/blob/eacc927378eed02af38966e9e2b50a731ce67452/tests/acceptance/test_snapshot.py).

@bahaa-ghazal From the two details above, for the next time I would appreciate if a member from the Mender Client team gives you an approval too - @alfrunes is an absolute ninja in golang, but there are some "Client team stuff", like the acceptance tests, that it will be better to get guidance from me or others in the client team :innocent: 

> This behavior looks undesirable, although I believe it was there for historical purposes

Yes and no. It has a historical meaning, but also is used in today's code because on an Open Source server the `deployments/next` API call does not take device's provides, only the artifact name, and for that reason the client (and mender-artifact tool) handle the two things differently.

PR #670, opened 2025-02-10
fix: Improve the error message when verifying ECDSA keys supplied via…

jo-lund:
… PKCS#11

When validating signatures with ECDSA keys exported from HSMs where the key is not P-256, i.e. the signature field length is not the correct length for P-256, inform that only ECDSA P-256 is supported.

Ticket: MEN-7941
Changelog: Title

### DIFF
--- a/artifact/signer.go
+++ b/artifact/signer.go
@@ -180,7 +180,8 @@ func UnmarshalECDSASignature(sig []byte) (r, s *big.Int, e error) {
 		return UnmarshalECDSASignatureASN1(sig)
 	}
 
-	return nil, nil, errors.Errorf("signer: invalid ecdsa key size: %d", len(sig))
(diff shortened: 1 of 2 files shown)


lluiscampos (approved):
Formatting and unit tests need an update, but otherwise:

It looks good to me! IMO this is enough information for a user to diagnose the incompatibility. Pre-approving.

It would be good with the eyes of @estenberg on this one - but if he takes too long to reply feel free to merge, @jo-lund .

jo-lund:

estenberg (approved):
But we already had an error message here? I wonder why this wasn't as the issue caught before then? Or did you already add this in a different PR?

Nitpick: I would say "unsupported signature length" (invalid sounds a bit harsh - it may still be a valid signature).

PR #675, opened 2025-02-18
fix: Correct the signature length check for ECDSA keys supplied via P…

jo-lund:
…KCS#11

An ASN.1 DER encoded signature is specified by RFC 3279 as: Ecdsa-Sig-Value ::= SEQUENCE  {
    r     INTEGER,
    s     INTEGER
}

ASN.1 gives an overhead of 6 bytes (SEQUENCE and INTEGER tags + length) and for P-256 "r" and "s" are 32 bytes each. This gives a minimum length of 70 bytes. "r" and "s" should be positive numbers so if the numbers are negative (msb is 1), the numbers need to be padded, i.e. they will use 33 bytes instead of 32. This means the encoded signature can be 71 or 72 bytes depending on padding.

Ticket: MEN-7941
Changelog: Title

### DIFF
--- a/artifact/signer.go
+++ b/artifact/signer.go
@@ -78,7 +78,8 @@ func (r *RSA) Verify(message, sig []byte, key interface{}) error {
 // ECDSA Crypto interface implementation
 const ecdsa256curveBits = 256
 const ecdsa256keySize = 32
-const ecdsa256Asn1SizeBytes = 72
+const ecdsa256Asn1SizeBytes = 70
+const ecdsa256Asn1Padding = 2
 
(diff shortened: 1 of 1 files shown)


jo-lund:

lluiscampos:
The failing integration test `TestRemoteTerminalOpenSource::test_bogus_proto_message` is broken in main pipelines, so it cannot be related to this PR.

Merging on your behalf :+1:

nickanderson:
@lluiscampos @jo-lund just wanted to highlight that @MuchoLucho noticed the commit wasn't signed
image

PR #677, opened 2025-02-22
MEN-8141: Require golang 1.22

lluiscampos:
Changelog: `mender-artifact` requires now go version 1.22

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -200,8 +200,9 @@ test:unit:mac:
     - mac-runner
 
 # Test that we can build with the golang version of the oldest supported yocto LTS release
+# MEN-8141: We don't support yocto kirkstone due to libraries requirements for new features
 test:backwards-compatibility:
-  image: golang:1.17.13-bullseye
+  image: golang:1.22.12-bullseye
   needs: []
   before_script:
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -200,8 +200,9 @@ test:unit:mac:
     - mac-runner
 
 # Test that we can build with the golang version of the oldest supported yocto LTS release
```
alfrunes, line 202: I suppose this line is invalid now as Kirkstone is still supported, right?
```diff
+ # Test that we can build with the golang version of the oldest supported yocto LTS release
- # Test that we can build with the golang version of the newest supported yocto LTS release
```

alfrunes (approved):

.gitlab-ci.yml
lluiscampos: I wanted to keep it, as to mark that supporting the oldest is the policy and we make an exception in this repo at this time due to the mentioned ticket (next line).

If you find it confusing I would remove the whole line - just modifying from "oldest" to "newest" I don't want to do because that can lead to confusion in one year from now.

danielskinstad (approved):

jo-lund (approved):

PR #687, opened 2025-03-18
chore: omit nil check

danielskinstad:
The static tests complains:
`should omit nil check; len() for nil slices is defined as zero (gosimple)`

https://gitlab.com/Northern.tech/Mender/mender-artifact/-/jobs/9442633456


lluiscampos (approved):

danielskinstad:

PR #697, opened 2025-04-23
Fix test-mac.sh

vpodzime:


mzedel (approved):
LGTM, thanks!

When we eventually switch to the more widely used commitlint I would say the `fix` commit would rather have to have a `test` type (as the test fix would otherwise bump the version as per [here](https://www.conventionalcommits.org/en/v1.0.0/#summary))

PR #701, opened 2025-05-13
fix: Mender-artifact fails on calling mender-update show-provides

michalkopczan:
Ticket: MEN-8382
Changelog: None

### DIFF
--- a/cli/util/ssh.go
+++ b/cli/util/ssh.go
@@ -1,0 +1,189 @@
+// Copyright 2025 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


danielskinstad (changes requested):
Well done, I think the code itself looks good. I didn't review the refactored code that closely, but checked that its been moved correctly and that it behaves the same. 

Regarding the commits, I agree with the previous discussion on separating the commits into two; one refac/chore and one fix. I also think this fix requires  a changelog. Etiher you could make the title (fix: bla bla bla) more explicit and set `Changelog: Title` or add a separate changelog entry describing it there

vpodzime:
Superseded by #703

PR #703, opened 2025-05-15
Men 8382 mender artifact fails on calling mender update show provides

michalkopczan:
New PR, disregard the old one (701)
Added two separate commits, one for the fix, another for ssh refactoring.
Compared to previous version, I removed fixes for restoring echo and hanging ssh in case remote device fails silently (e.g. power loss). Will move it to new tasks.

### DIFF
--- a/cli/util/ssh.go
+++ b/cli/util/ssh.go
@@ -1,0 +1,183 @@
+// Copyright 2025 Northern.tech AS
+//
+//    Licensed under the Apache License, Version 2.0 (the "License");
+//    you may not use this file except in compliance with the License.
+//    You may obtain a copy of the License at
+//
+//        http://www.apache.org/licenses/LICENSE-2.0
+//
+//    Unless required by applicable law or agreed to in writing, software
+//    distributed under the License is distributed on an "AS IS" BASIS,
+//    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//    See the License for the specific language governing permissions and
+//    limitations under the License.
+
+package util
+
+import (
+	"bufio"
+	"context"
+	"io"
+	"os"
+	"os/exec"
+	"os/signal"
+	"strings"
+	"syscall"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+type SSHCommand struct {
+	Cmd     *exec.Cmd
+	ctx     context.Context
+	Stdout  io.ReadCloser
+	cancel  context.CancelFunc
+	sigChan chan os.Signal
+	errChan chan error
+}
+
+func StartSSHCommand(c *cli.Context,
+	_ctx context.Context,
+	cancel context.CancelFunc,
+	command string,
+	sshConnectedToken string,
+) (*SSHCommand, error) {
+
+	var userAtHost string
+	var sigChan chan os.Signal
+	var errChan chan error
+	port := "22"
+	host := strings.TrimPrefix(c.String("file"), "ssh://")
+
+	if remotePort := strings.Split(host, ":"); len(remotePort) == 2 {
+		port = remotePort[1]
+		userAtHost = remotePort[0]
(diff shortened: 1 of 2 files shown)


cli/write.go
```diff
@@ -1019,27 +1026,66 @@ func showProvides(c *cli.Context) (map[string]string, error) {
...
+	}
+
+	if err := cmd.Start(); err != nil {
 		return nil, err
 	}
-	if len(stdout) == 0 {
-		return nil, nil
+
+	// Wait for 60 seconds for ssh to establish connection
+	err = waitForBufferSignal(stdout, os.Stdout, sshInitMagic, 2*time.Minute)
```
vpodzime: Looks like 2 minutes and thus 120 seconds to me.

vpodzime (commented):
Looks generally good to me, but I cannot assess the quality of the Go code due to lack of knowledge/experience. @alfrunes please help. :)

danielskinstad (commented):
Again, I think the code looks good :)

This is perhaps a personal preference, feel free to dismiss, but personally I would prefer a bit shorter commit titles and rather do something like:
`fix: call show-provides with sudo`
`Changelog: Fix crash when generating artifact from ssh by adding sudo to show-provides and properly handle sudo asking for password`

cli/write.go
```diff
@@ -1019,27 +1026,66 @@ func showProvides(c *cli.Context) (map[string]string, error) {
...
+	if err == nil {
+		sigChan = make(chan os.Signal, 1)
+		errChan = make(chan error, 1)
+		// Make sure that echo is enabled if the process gets
+		// interrupted
+		signal.Notify(sigChan)
+		go util.EchoSigHandler(ctx, sigChan, errChan, term)
+	} else if err != syscall.ENOTTY {
+		return nil, err
+	}
```
alfrunes: Are we intentionally ignoring all errors except ENOTTY?

cli/util/ssh.go
```diff
@@ -0,0 +1,183 @@
...
+	"syscall"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/urfave/cli"
+)
+
+type SSHCommand struct {
+	Cmd     *exec.Cmd
+	ctx     context.Context
```
alfrunes, line 34: The `ctx` is unused.
```suggestion
```
Side-note; storing Context in structs is in most cases considered an [anti-pattern](https://go.dev/blog/context-and-structs#storing-context-in-structs-leads-to-confusion) in Go

cli/util/ssh.go
```diff
@@ -0,0 +1,183 @@
...
+				break
+			}
+		}
+	}()
+
+	select {
+	case err = <-errChan:
+		// Error from goroutine
+	case <-time.After(deadline):
+		err = errors.New("Input deadline exceeded")
```
alfrunes, line 180: We could also reuse the deadline from the context in the parent scope here instead of a fixed duration.
```suggestion
	case <-ctx.Done():
		err = ctx.Err()
```

alfrunes (approved):

cli/util/ssh.go
alfrunes: Correct. As long as the heap reference is reachable, the garbage collector should not touch it.

cli/util/ssh.go
alfrunes: Sorry for being unclear. My point was to not have multiple sources of deadlines. The context itself can be used to provide a deadline (see [context.WithTimeout](https://pkg.go.dev/context#WithTimeout)). In the above select case we allocate a new timer and ignore the context which is used in parallel by other go routines.

michalkopczan:

cli/util/ssh.go
vpodzime: So the conclusion is that it should be dropped?

michalkopczan: I'm going to merge it as-is. It works.

PR #710, opened 2025-06-04
ci: Stop publishing MacOS/darwin binary

vpodzime:
mender-artifact is now available in brew and we want users to use that instead of binaries. The documentation no longer has links to the binaries so we can stop publishing them.

Ticket: MEN-8354
Changelog: none


danielskinstad (approved):
LGTM

PR #713, opened 2025-06-12
RFC - feat: let macos also process sdimages by relying on fdisk for partitioning

mzedel:


.gitlab-ci.yml
```diff
@@ -38,19 +38,18 @@ variables:
...
-  -  sleep "${DOCKER_RETRY_SLEEP_S}"
+  - docker ps && DOCKER_RUNNING=true
+  - if [ "${DOCKER_RUNNING}" == "true" ]; then
+  - break
+  - fi
+  - sleep "${DOCKER_RETRY_SLEEP_S}"
   - done
   - if [ "${DOCKER_RUNNING}" != "true" ]; then
-  -  exit 192
+  - exit 192
```
vpodzime, line 52: Unrelated changes.

vpodzime (changes requested):
A slightly better than what I expected, but exactly what I assumed -- 50% wrong. And that's not really helpful. I spent as much time reviewing this and commenting on the issues as I would making the required changes. That would actually work. But a nice try for sure!

PR #717, opened 2025-06-18
chore: Look for binaries in brew specific path

elkoniu:
When brew installs binary it will create symlink directory in location like `/usr/local/opt/<binary>/bin`. This location is not part of the system PATH by default.

When mender-artifact evaluates commands location it need to check this brew specific directories apart of default ones.

ticket: MEN-8471
changelog: Extend tools search locations with brew specific ones

### DIFF
--- a/utils/binpath.go
+++ b/utils/binpath.go
@@ -16,6 +16,7 @@ package utils
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
 	"path"
 	"runtime"
@@ -26,6 +27,10 @@ import (
 
 var (
 	ExternalBinaryPaths = []string{"/usr/sbin", "/sbin", "/usr/local/sbin"}
+)
+
+var (
+	BrewSpecificPaths = []string{"/usr/local/opt"}
 )
 
 var unsupportedBinariesDarwin = []string{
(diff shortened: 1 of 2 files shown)


utils/binpath.go
```diff
@@ -51,6 +56,18 @@ func GetBinaryPath(command string) (string, error) {
 		}
 	}
 
+	// look for the binaries in brew symlink directories
+	// example: /usr/local/opt/e2fsprogs/bin, /usr/local/opt/mtools/bin etc.
+	for _, p = range BrewSpecificPaths {
+		items, err := os.ReadDir(p)
```
vpodzime, line 62: We need to skip files in case there are some.

utils/binpath.go
```diff
@@ -28,6 +29,10 @@ var (
 	ExternalBinaryPaths = []string{"/usr/sbin", "/sbin", "/usr/local/sbin"}
 )
 
+var (
+	BrewSpecificPaths = [string]{"/usr/local/opt"}
```
vpodzime, line 33: Not sure if this `[string]` syntax is also acceptable, but `[]string` seems to be the right or at least preferred one.

utils/binpath.go
```diff
@@ -51,6 +56,18 @@ func GetBinaryPath(command string) (string, error) {
 		}
 	}
 
+	// look for the binaries in brew symlink directories
+	// example: /usr/local/opt/e2fsprogs/bin, /usr/local/opt/mtools/bin etc.
+	for _, p = range BrewSpecificPaths {
+		items, err := os.ReadDir(p)
+		for _, d := range items {
+			p, err = exec.LookPath(path.Join(d, "bin",  command))
```
vpodzime, line 64: `e2fsprogs`, for example, have some binaries in `sbin` so we need to check that too.

utils/binpath.go
elkoniu: I added the check to skip all non-directory items.

utils/binpath.go
elkoniu: In the function 1st it will check in PATH, then in `ExternalBinaryPaths` which covers `/sbin`, then it will evaluate `/usr/ocal/opt`. So in case of `e2fsprogs` it should pick it up from `ExternalBinaryPaths` existing already.

elkoniu:

elkoniu:

PR #722, opened 2025-07-16
chore: Add help info when no make target specified

elkoniu:
In the documentation addressing `build from source` we suggest users to follow
repository instructions for each repo. `mender-artifact` is based on plain Makefile
and when `make` is called without extra parameters build on host will be performed.

In README we suggest to use `make build-natives-contained` as a main build command.
This is not what Makefile is doing by default.

My proposition is to introduce `help` target with displaying some most popular targets,
but not performing any (potential harmful) action by default.

This way user will need to explicite tell `make` what action needs to be performed. 

Ticket: None
Changelog: Add help info to Makefile targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,6 +35,26 @@ ifneq ($(GOOS),linux)
 	TAGS += nopkcs11
 endif
 
+default: help
+
+help:
+	@echo "make [TARGETS...]"
+	@echo
+	@echo "The following targets are available:"
+	@echo
+	@echo "    help:               Print this usage information."
+	@echo
+	@echo "    build:              Build the code on the localhost"
+	@echo "    build-contained:    Build the code in the container"
+	@echo "    clean:              Remove build artifacts"
+	@echo
+	@echo "    install:            Install build artifacts on the locally"
+	@echo "    uninstall:          Uninstall build artifacts from the localhost"
+	@echo
+	@echo "    test:               Run tests"
+	@echo "    check:              Run various code checking tools"
+	@echo "    coverage:           Run code coverage checks"
+
 build:
 	$(GO) build $(GO_LDFLAGS) $(BUILDV) -tags '$(TAGS)'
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -35,6 +35,25 @@ ifneq ($(GOOS),linux)
 	TAGS += nopkcs11
 endif
 
+.PHONY: help
```
vpodzime, line 38: I believe it's enough to have this near the bottom of the file, not necessary to add noise here.

Makefile
```diff
@@ -35,6 +35,25 @@ ifneq ($(GOOS),linux)
...
+help:
+	@echo "make [TARGETS...]"
+	@echo
+	@echo "The following targets are available:"
+	@echo
+	@echo "    help:               Print this usage information."
+	@echo
+	@echo "    build:              Build the code on the localhost"
+	@echo "    build-containered:  Build the code in the container"
```
vpodzime, line 47: It's *build-contained* :wink:

vpodzime (commented):
Good idea in general! `make` does use some target by default so we could make sure that actually triggers the `build` target.

Makefile
elkoniu: copy-pasta error - fixed :)

vpodzime (approved):
Looks good to me

danielskinstad:
Just a questoin: shouldn't `build` be the default target here? The `Makefile Conventions` manual from GNU states that `all` should be the default target, which we don't have, so I guess the closest is `build`

vpodzime:
> Just a questoin: shouldn't `build` be the default target here?

That's a common behavior. But I don't have a strong opinion here because our targets are quite special and so telling users what their options are is nice.

vpodzime (commented):
I don't think we should/need to install `e2tools`. Also, *requites* in the second commit message.

elkoniu:
> I don't think we should/need to install `e2tools`. Also, _requites_ in the second commit message.

Spelling fixed - thanks for pointing this our. For the `e2tools` - I removed it and just forced `reinstall` on top of the main package.

PR #723, opened 2025-07-18
fix!: Add validation rules to artifact name

alfrunes:
Constrain the name to 256 characters and all characters must be in the graphical Unicode categories: L, M, N, P, S, Zs.

BREAKING CHANGE: Artifact name must be less than 256 characters and from
                 Unicode categories: L, M, N, P, S, Zs.


alfrunes:

PR #734, opened 2025-08-20
QA-1017: chore(CODEOWNERS): Remove global ownership, add dependency files

lluiscampos:
In the context of QA-1017, the original motivation was to move away from deprecated
dependabot reviewers but not necessarily enable code reviewers feature across all files. This commit
amends the original work so that only dependency update related PRs are responsibility of the
client-dependabot-reviewers team

Ticket: QA-1017

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,7 @@
-# Default owners for everything
-* @mendersoftware/client-dependabot-reviewers
+# Dependency updates
+go.mod @mendersoftware/client-dependabot-reviewers
+go.sum @mendersoftware/client-dependabot-reviewers
+Dockerfile* @mendersoftware/client-dependabot-reviewers
+*requirements*.txt @mendersoftware/client-dependabot-reviewers
+package.json @mendersoftware/client-dependabot-reviewers
+package-lock.json @mendersoftware/client-dependabot-reviewers


CODEOWNERS
```diff
@@ -1,2 +1,8 @@
-# Default owners for everything
-* @mendersoftware/client-dependabot-reviewers
+# Dependencies and build files
+go.mod @mendersoftware/client-dependabot-reviewers
+go.sum @mendersoftware/client-dependabot-reviewers
+Dockerfile* @mendersoftware/client-dependabot-reviewers
+requirements.txt @mendersoftware/client-dependabot-reviewers
+requirements*.txt @mendersoftware/client-dependabot-reviewers
+package.json @mendersoftware/client-dependabot-reviewers
+package-lock.json @mendersoftware/client-dependabot-reviewers
```
danielskinstad, line 8: We don't have these files here (except for in the vendor/ directory). We do have `apk-requirements.txt` and `deb-requirements.txt`, which I guess don't need to be trackesd by dependabot

mzedel (approved):
LGTM - let's hope it works as expected!

CODEOWNERS
lluiscampos: > We don't have these files here (except for in the vendor/ directory).

Yes, thanks for noticing. I was a bit back and forth on this but what I decided was to make a "generic" list for client repositories and then put the same CODEOWNERS file everywhere. That is why I include Python or javascript/node, to avoid "thinking" in every repo and potentially missing something. Besides it also makes it more future proofed :mag: 

> We do have `apk-requirements.txt` and `deb-requirements.txt`, which I guess don't need to be trackesd by dependabot

These are custom made by us for CI, so dependabot does not know about it.

danielskinstad: > a "generic" list for client repositories and then put the same CODEOWNERS file everywhere.

Alright, that makes sense. Should we exclude the vendor directory to ensure it doesn't try to do anything fishy?

lluiscampos: The vendor will be updated together with go.mod / go.sum, right? Besides this only would ping for reviewers, is not modifying dependabot configuration per se.

I'll leave this part as-is and see if some issue arises.

danielskinstad: > The vendor will be updated together with go.mod / go.sum, right?

Yes, I just meant that there's no reason to check the requirements.txt and similar files in the vendored folder, but we can wait and see

mzedel (approved):

mzedel (approved):

PR #753, opened 2025-10-06
build: Fix `make test` target

lluiscampos:
The well-intended commit ba669f03 broke `make test` when improving the `make coverage` target :)


Makefile
```diff
@@ -132,7 +132,7 @@ tooldep:
 	mtools --version
 
 test: tooldep
-	$(GO) test -v $(PKGS)
+	$(GO) test -v ./...
```
elkoniu, line 135: So `PKGS = $(shell go list ./... | tr '\n' ',' | head -c -1)` is just overcomplicated version of simple `./...` ?

elkoniu: Also `PKGS` seems to be not used anywhere else so I think it may be removed.

PR #836, opened 2026-07-14
fix(cli): resolve symlinks against their own directory when signing

lgrosz:
## What

`mender-artifact sign` fails on relative symlinks with `Could not get artifact file stat` whenever the process working directory is not the symlink's directory. This PR replaces the manual symlink handling in `signExisting()` with a single `filepath.EvalSymlinks` call on the input path.

## The bug

`signExisting()` resolved symlinks in userspace and got the resolution base wrong:

```go
artFile, err = os.Readlink(artFile) // raw link text, e.g. "core-image-...-20260714.mender"
artFileStat, err = os.Stat(artFile) // resolved against the process cwd, not the link's dir
```

`os.Readlink` returns the target exactly as stored in the link. When that target is relative, the subsequent `os.Stat` resolves it against the process working directory instead of the symlink's own directory.

### Example: signing a Yocto-built artifact

Yocto/meta-mender deploys artifacts as a versioned file plus a **relative** symlink:

```
$ ls -l tmp/deploy/images/raspberrypi4/
core-image-full-cmdline-raspberrypi4.mender -> core-image-full-cmdline-raspberrypi4-20260714120000.mender
core-image-full-cmdline-raspberrypi4-20260714120000.mender
```

Signing via the symlink from anywhere other than that directory fails:

```
$ cd ~/build   # anywhere that isn't the images/raspberrypi4 directory
$ mender-artifact sign \
      -k private.key \
      tmp/deploy/images/raspberrypi4/core-image-full-cmdline-raspberrypi4.mender
Could not get artifact file stat
```

`mender-artifact read` on the same path works. `mender artifact sign` works when running from the pwd of the symlink. The current workaround would be to canonicalise the path first, and sign that path instead.

## The fix

Resolve the input path once with `filepath.EvalSymlinks`.

Fixes that fall out of this:

* `-o` / `--output-path` is no longer silently overridden when the input is a symlink
* The temporary file is created in the resolved target's directory rather than the link's, so the final `os.Rename` stays on one filesystem.
* Chains of symlinks now work

## Tests

* `TestSignViaRelativeSymlink` — regression test for the scenario above: relative link target, cwd elsewhere. Fails with `Could not get artifact file stat` on the old code.
* `TestSignViaSymlinkChain` — a two-hop chain of relative symlinks.
* `TestSignViaSymlinkWithOutputPath` — `--output-path` is honoured for symlink inputs and the original artifact is left alone.

The existing `TestSignViaSymlink` (absolute link target) still passes and is kept as-is.

Ticket: None

### DIFF
--- a/cli/sign.go
+++ b/cli/sign.go
@@ -40,6 +40,11 @@ func signExisting(c *cli.Context) error {
 	}
 
 	artFile := c.Args().First()
+	artFile, err = filepath.EvalSymlinks(artFile)
+	if err != nil {
+		err = errors.Wrapf(err, "Can not resolve: %s", c.Args().First())
+		return cli.NewExitError(err, 1)
+	}
 	outputFile := artFile
 	if len(c.String("output-path")) > 0 {
 		outputFile = c.String("output-path")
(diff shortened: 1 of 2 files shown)


michalkopczan:
Thank you @lgrosz for your contribution! Looks good, nice fix, will merge as soon as the test pipeline finishes.

michalkopczan:

lgrosz:
> Thank you @lgrosz for your contribution! Looks good, nice fix, will merge as soon as the test pipeline finishes.

Thank you!

michalkopczan (approved):

PR #846, opened 2026-08-05
fix(cli): Correct binary name in no-argument error messages

lluiscampos:
Fixes the usage hint of read, validate, sign and modify commands ("mender-artifact" instead of "artifacts"), and the suggested subcommand in modify ("modify" instead of "read").

Assisted-by: OpenCode:claude-sonnet-4.6

### DIFF
--- a/cli/modify.go
+++ b/cli/modify.go
@@ -42,7 +42,7 @@ func modifyArtifact(c *cli.Context) (err error) {
 
 	if c.NArg() == 0 {
 		return cli.NewExitError("Nothing specified, nothing will be modified. \n"+
-			"Maybe you wanted to say 'artifacts read <pathspec>'?", 1)
+			"Maybe you wanted to say 'mender-artifact modify <pathspec>'?", 1)
 	}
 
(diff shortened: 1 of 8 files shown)


lluiscampos:
Hello Mender team 👋 

This PR is the result of me experimenting with a new agent workflow. In other words it does **not fix something that I found** but something that the agent found.

Let me know if you rather don't want these kind of "improvements" and I won't submit them in the future. No hard feelings.

Cheers 🍻

lluiscampos:
@michalkopczan @danielskinstad Please just acknowledge that we don't have capacity for minor fixes like this one and I'll just drop it. To be honest it is quite demotivating coming here regularly to see nothing happening 😢

danielskinstad (approved):
Hey, @lluiscampos! This is something we should have capacity for, so I would say we definitely want changes like this. I believe this is a case of things falling in between the cracks. Sorry for not getting back to you sooner!

danielskinstad:

danielskinstad:
We have some issues with the integration tests atm. I will get this merged once they are green.

lluiscampos:
@danielskinstad Thank you very much 😃

PR #847, opened 2026-08-12
Feat/delta image generation

mzedel:
This should follow the documented behaviour of the delta image generator, but avoids both the build complexity of including it and instead relies on the external packages and removes the need to go through another tool to end up with a (mender-)artifact.


vpodzime:
Looks good to me, generally, though my knowledge of golang is too limited for me to be able to review this properly. **However**, it goes against the existing pattern of Artifact generators being tied to the respective Update Modules. The `rootfs-image` type is an exception for historical reasons, but other UMs have their respective generators utilizing the generic `mender-artifact` "archiver/packager/...".

PR #848, opened 2026-08-14
feat(artifact): honor SOURCE_DATE_EPOCH for reproducible artifacts

voxeljorge:
## Problem

`mender-artifact write` is not byte-reproducible. Two runs over identical inputs, seconds apart, produce artifacts differing in exactly 7 bytes.

`FileArchiver.Write` stats the file it is archiving and builds the tar header with `tar.FileInfoHeader`. The files reaching this path are temporaries created during the run, so the header inherits an mtime that changes on every invocation and, on Linux, the uid/gid of whoever ran the build. The differing bytes land on the mtime (136–147) and header checksum (148–155) fields of the `header.tar.gz` and `data/0000.tar.gz` entries.

This appears to be the only source of non-determinism. `version` and `manifest` are already byte-identical between runs, because `StreamArchiver.Write` and the manifest signer in `awriter/signer.go` construct their headers manually with only `Name`, `Mode` and `Size` — leaving ModTime and ownership zero. The contrast is visible within a single artifact:

```
-rw------- 0/0           31  1970-01-01 00:00  version
-rw------- 0/0          242  1970-01-01 00:00  manifest
-rw------- user/group   350  2026-08-14 20:49  header.tar.gz
-rw------- user/group   129  2026-08-14 20:49  data/0000.tar.gz
```

## Change

Honor [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/). When it is set, pin the header mtime to it and zero the ownership fields. When it is unset, output is unchanged.

The zero ownership is not a new policy — it is what `StreamArchiver` and the signer already emit, so this brings the third header-writing path in line with the other two.

This is deliberately opt-in. `FileArchiver` is also used to archive the payload files themselves, whose mtimes come from the user's real input files and are surfaced by `mender-artifact read` as the `modified` field. Zeroing unconditionally would discard that, so the variable is what distinguishes "I want a reproducible build" from normal use.

## On failing loudly for a malformed value

A malformed `SOURCE_DATE_EPOCH` returns an error rather than being silently ignored. This is intentional, and I would argue it is the safer default.

The variable exists so that a caller can assert reproducibility. If the value is junk and the build quietly falls back to wall-clock mtimes, it still succeeds and still appears reproducible while silently not being — and that stays invisible until someone diffs two artifacts and has to work out why they differ. A loud failure at the point of the mistake is much cheaper than a silent one discovered later, possibly by someone else. The SOURCE_DATE_EPOCH specification also recommends exiting non-zero on a malformed value.

That said, it does introduce a failure mode that did not exist before, so if you would rather this warn or silently ignore, I am happy to change it.

## Verification

Same inputs, two builds two seconds apart:

```
with SOURCE_DATE_EPOCH set:  b9551caa…   b9551caa…    identical
without:                     0086b520…   259cea6d…    7 bytes differ
```

## Tests

Three added to `artifact/tar_writer_file_test.go`:

- identical output across differing source mtimes when the variable is set
- default behaviour preserving the file mtime when it is unset
- an error on a malformed value

The full suite passes, with the exception of `cli.TestCopyRootfsImage` and `cli.TestCopyFromStdin`, which fail identically on unmodified `master` in my environment (they need rootfs image tooling I do not have available) and are unrelated to this change.

## Possible follow-up

If you would prefer artifacts to be reproducible by default rather than opt-in, the narrower fix would be to zero mtime and ownership only where `FileArchiver` wraps the internally-created temporaries, leaving payload archiving untouched. That requires distinguishing the two roles inside `awriter/writer.go`, so I have not assumed it here — happy to follow up if that is the direction you would rather take.

### DIFF
--- a/artifact/tar_writer_file.go
+++ b/artifact/tar_writer_file.go
@@ -18,9 +18,33 @@ import (
 	"archive/tar"
 	"io"
 	"os"
+	"strconv"
+	"strings"
+	"time"
 
 	"github.com/pkg/errors"
 )
+
+// sourceDateEpochVar is the reproducible-builds convention for pinning
+// timestamps embedded in build output.
+// See https://reproducible-builds.org/docs/source-date-epoch/.
+const sourceDateEpochVar = "SOURCE_DATE_EPOCH"
+
+// sourceDateEpoch reports the timestamp requested by SOURCE_DATE_EPOCH, and
+// whether the variable was set at all. A malformed value is an error rather
+// than a silent fallback: a build that believes it is reproducible and is not
+// is worse than one that stops.
+func sourceDateEpoch() (time.Time, bool, error) {
+	v, ok := os.LookupEnv(sourceDateEpochVar)
+	if !ok {
+		return time.Time{}, false, nil
+	}
+	secs, err := strconv.ParseInt(strings.TrimSpace(v), 10, 64)
+	if err != nil {
+		return time.Time{}, false, errors.Wrapf(err, "arch: invalid %s value %q", sourceDateEpochVar, v)
+	}
+	return time.Unix(secs, 0).UTC(), true, nil
+}
 
 type FileArchiver struct {
 	*tar.Writer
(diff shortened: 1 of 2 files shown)


voxeljorge:
I see that there are some gitlab failures but they look like they might be unrelated to this PR.

voxeljorge:
Hi @michalkopczan I see that you assigned this, please let me know if you have any feedback. We're using this change in our fork but would like to avoid using a forked version if possible

michalkopczan:
Hi @voxeljorge

Here's a task used for tracking progress: https://northerntech.atlassian.net/browse/MEN-10162

Thank you for your contribution.

In `func (fa *FileArchiver) Write(f *os.File, archivePath string) error {` you mention that "files reaching this writer are temporaries created during the current run" but it is not true in all cases. This function is also used to write the payload files of the artifact (contents of data/0000.tar.gz). Using the env var to reset file ownership results in the payload files to have their owner/group being set to root.

This is fine for reproducible builds, but changing the ownership of the payload by setting SOURCE_DATE_EPOCH may lead to unintended consequences. As in: someone want's to have reproducible builds with regard to mtime, sets SOURCE_DATE_EPOCH, and thus unintentionally makes payload files accessible by unintended users.

The difference between manifest/version ownership and payload's ownership is that manifest/version files are used exclusively by mender-update, while the payload is not.

Have you considered using a separate variable for ownership resetting? Alternatively we could go into the direction of resetting ownership of header.tar.gz + 0000.tar.gz by default while leaving the contents of 0000.tar.gz (the payload) with the original owner+mtime as it is now.

voxeljorge:
Thanks @michalkopczan. You're right that `FileArchiver` also writes the payload files, and the comment saying otherwise was wrong.

Agreed that ownership is out of scope for `SOURCE_DATE_EPOCH`; the spec only covers timestamps. I've pushed a change that drops the ownership reset, so the variable now only pins mtimes, and fixes the comment. If we need a separate variable for ownership we'll open a separate PR for it with a proposed solution, but I believe we can handle ownership in the build system.

PR #853, opened 2026-09-02
fix: Handle artifacts with empty type-info in headers gracefully

vpodzime:
Although such artifacts don't strictly adhere to the format
specification, past versions of mender-artifact itself produced
such artifacts. Rejecting them as invalid/corrupt causes issues
in many places because it's very hard to recreate all past
artifacts with the latest version of mender-artifact.

Artifacts with an actual mismatch between the two type-info
fields are still rejected because that's an indication of a real
issue with Artifact creation as no past versions of
mender-artifact have done that.

Ticket: MEN-9561
Changelog: Artifacts produced by past versions of mender-artifact
missing type-info in headers are no longer rejected as
invalid/corrupt
Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Signed-off-by: Vratislav Podzimek <vratislav.podzimek+auto-signed@northern.tech>

### DIFF
--- a/areader/reader.go
+++ b/areader/reader.go
@@ -48,6 +48,7 @@ type Reader struct {
 	ForbidUnknownHandlers     bool
 
 	warnOnValidationErrors bool
+	warnOnEmptyTypeInfo    bool
 	logger                 utils.Logger
 
 	shouldBeSigned  bool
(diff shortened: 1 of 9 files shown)


alfrunes (approved):

danielskinstad (commented):
Should we also handle validate and other commands?
```
daniel@daniel-debian:~/mender-repos/mender-artifact$ ./mender-artifact validate old-rootfs.mender
readHeaderV3: handleHeaderReads: reader: can not read header: Type in type-info header is empty: This is a known bug in some versions of mender-artifact prior to 4.4.0. Please recreate the artifact with version 4.4.0 or newer.
```

vpodzime:
> Should we also handle validate and other commands?

That's a good question. I would vote for `validate` to fail with an error. But the real-world use case of validating signatures and checksums/... in old artifacts suggests otherwise.

danielskinstad (approved):
Looks good to me

vpodzime:

alfrunes (approved):

vpodzime:
* 4.4.x

lgrosz:
Thank you! I just ran into this when upgrading `mender-artifact` on my CI pipeline that handled artifacts generated from an older version contained within a Yocto project.

PR #856, opened 2026-09-15
chore: only propose go modules compatible with go 1.22

danielskinstad:
Renovate's strict constraints filtering skips module versions whose go directive is above 1.22, so we stay buildable with the Go from our oldest supported Yocto LTS. The constraint has to be an explicit range: the one Renovate derives from go.mod (~1.22.x) also filters out releases with a lower go directive, which drops nearly every update. Security PRs opt out of the filter so a fix never gets dropped silently.

Ticket: None


vpodzime (approved):
I'd do `chore(renovate)`. And I can't say I fully understand the commit message, but the changes look reasonable.

michalkopczan (approved):

